PR #44, opened 2025-02-12
Development Update

paullizer:
Add Content Safety to the service as an optional capability 
Add AI Search and Doc Intelligence to Admin Settings
Add Your documents and Groups as an optional capability
Track AI Model Used for Generating Responses


paullizer (commented):
Updating to v196.9

PR #129, opened 2025-04-29
Added User selected model and Video Indexer fields and backend logic to admin settings

paullizer:


paullizer:
need to update group workspace to support video, merge, then we can delete the branch

PR #146, opened 2025-05-05
Bug creating too many search returns

paullizer:


paullizer:
Resolving bug that was returning too many results with the legacy chunk size.

PR #216, opened 2025-05-22
Suggested update to README.md

ChrisSheppard-UICGS:
In deploying the latest release, I found that uploading documents to a user workspace fails when Enhanced Citations are enabled but containers have not manually been added to the storage account.


paullizer:
Good catch! I will update the app logic to create the containers if they are missing.

PR #220, opened 2025-05-23
Fix route_backend_settings uncontrolled data used in path expression

odaysec:
Accessing files using paths constructed from user-controlled data can allow an attacker to access unexpected resources. This can result in sensitive information being revealed or deleted, or an attacker being able to influence behavior by modifying unexpected files.

https://github.com/microsoft/simplechat/blob/c4d0d4ebf555427ce2b74fec34dcc5e5de2a8ee1/application/single_app/route_backend_settings.py#L17-L19

https://github.com/microsoft/simplechat/blob/c4d0d4ebf555427ce2b74fec34dcc5e5de2a8ee1/application/single_app/route_backend_settings.py#L41-L43

Fix the issue need to validate and sanitize the user-provided input (`idx_type`) before using it to construct the file path. The best approach is to:
1. Use `os.path.normpath` to normalize the constructed path and remove any `..` segments.
2. Ensure that the normalized path is contained within the intended directory (`static/json`) by checking that it starts with the base directory path.
3. Optionally, use `werkzeug.utils.secure_filename` to sanitize the filename component (`ai_search-index-{idx_type}.json`) to ensure it does not contain any special characters.

The changes will be made in the `fix_index_fields` and `check_index_fields` functions in `application/single_app/route_backend_settings.py`.

---

#### References
[Path Traversal](https://owasp.org/www-community/attacks/Path_Traversal)
[werkzeug.utils.secure_filename](http://werkzeug.pocoo.org/docs/utils/#werkzeug.utils.secure_filename)

### DIFF
--- a/application/single_app/route_backend_settings.py
+++ b/application/single_app/route_backend_settings.py
@@ -14,9 +14,12 @@ def register_route_backend_settings(app):
         idx_type = data.get('indexType')  # 'user' or 'group'
 
         # load your golden JSON
-        fname = f'ai_search-index-{idx_type}.json'
-        fpath = os.path.join(current_app.root_path, 'static', 'json', fname)
-        with open(fpath,'r') as f:
+        fname = secure_filename(f'ai_search-index-{idx_type}.json')
+        base_path = os.path.join(current_app.root_path, 'static', 'json')
+        fpath = os.path.normpath(os.path.join(base_path, fname))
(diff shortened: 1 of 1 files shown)


paullizer:
I like this! Great feedback. I will test it out, due to the holiday weekend, it will be next week. I really appreciate sharing your knowledge and expertise to improve the security and functionality!

PR #412, opened 2025-09-05
Feature - Added Managed Identity authentication option to Enhanced Citation Storage Account + bug fixes

Xeelee33:
**Feature - Added Managed Identity authentication option to Enhanced Citations Storage Account Admin Setting**
_Current Enhanced Citation Setting_
image

_Updated Enhanced Citation Setting_
Added dropdown selection
image

Connection String
image

Managed Identity
image

**Bugfix 1 - Removed duplicate logo_version setting**
[Line 389 in route_frontend_admin_settings.py](https://github.com/microsoft/simplechat/blob/1c93e7f4a3d2401143ac69de0506c8f87b06cb5c/application/single_app/route_frontend_admin_settings.py#L389)
image

**Bugfix 2 - Fixed Video Indexer settings in Admin Settings Setup Walkthrough so that API Key is optional since Video Indexer requires ARM authentication now instead of API Key auth**
_Current_
Step 8 requires Location, Account ID, and API Key to move to next step (it says Endpoint is required, but the actual case statement only checks for Location, Account ID, and API Key). However, Video Indexer service now requires authentication via Entra ID/ARM and doesn't even provide the option to connect with API key, so this step is impossible to complete for new and recent deployments.
image

_Updated_
Changed API Key to Optional and updated the case statement logic to require an Endpoint, Location, and Account ID, all of which are required to initialize client via ARM.
image

### DIFF
--- a/application/single_app/config.py
+++ b/application/single_app/config.py
@@ -611,11 +611,14 @@ def initialize_clients(settings):
 
         try:
             if enable_enhanced_citations:
-                blob_service_client = BlobServiceClient.from_connection_string(settings.get("office_docs_storage_account_url"))
-                CLIENTS["storage_account_office_docs_client"] = blob_service_client
-                
-                # Create containers if they don't exist
-                # This addresses the issue where the application assumes containers exist
+                if settings.get("office_docs_authentication_type") == "key":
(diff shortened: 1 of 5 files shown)


paullizer:
Great update!

PR #511, opened 2025-11-11
Cosmos DB Scripts for Managed Identity Data Plane Permissions, and Database and Container Creation

jarredm-demo:
Using a Managed Identity is the safest method of deploying.  However, a Managed Identity can only access the Data Plane inside Cosmos DB, and in order to do so these permissions must be set via script (supplied Azure CLI and Azure PowerShell in this branch).  Since Database and Container creation for Azure Cosmos DB is a Control Plane operation, these also need to be scripted and performed directly during deployment, as calls to the REST API go through the Data Plane and therefore cannot use the Managed Identity (they require Keys).  Eventually the project aims to avoid using keys were possible.


jarredm-demo:
@microsoft-github-policy-service agree company="Microsoft"

I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.

PR #551, opened 2025-12-19
Bicep support for 'custom' azure environment

clarked-msft:
- Set CUSTOM_ app env variables
- Use az.environment() to determine cloud
- Add path to user search graph url when using 'custom'
- Give video_indexer_endpoint a default value in 'custom' since is required
- Expose bicep param to set openai model params for existing OpenAI
- Fix environment variable name, AZURE_ENVIRONMENT

### DIFF
--- a/docs/explanation/release_notes.md
+++ b/docs/explanation/release_notes.md
@@ -21,7 +21,19 @@
     *   **Removed Duplicate Comment**: Cleaned up duplicate "Render user-search results" comment.
     *   **Impact**: Member management buttons now render and function correctly, provide better error feedback, and auto-recover from stale member data.
     *   (Ref: `manage_group.js`, event handler deduplication, error handling improvements, toast notifications)
-    
+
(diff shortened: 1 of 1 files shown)


paullizer:
@clarked-msft , @SteveCInVA submitted a huge PR that may have made this OBE. please review. 

we will be using azd moving forward and the bicep it uses, please review.

clarked-msft:
> @clarked-msft , @SteveCInVA submitted a huge PR that may have made this OBE. please review.
> 
> we will be using azd moving forward and the bicep it uses, please review.

These changes should be to the latest bicep Steve submitted.  My understanding is azd does not work AGTS but that is fine.  Setup is very minimal from the bicep.

clarked-msft:
Latest bicep changes are merged.  Take a look at the 'cloudEnvironment' param and make sure you're okay with this...  The cloud name is already available through az.environment() so I changed this to be the value SimpleChat uses and is really only needed to force 'custom' mode for testing.

PR #562, opened 2026-01-06
Add flask instrumentation

clarked-msft:
Flask instrumentation with option to disable in case of telemetry issues.  This surfaces flask requests in app insights with an operation ID, and links traces, exceptions etc using operation id.

Down the road the direct app insights logging like log_event from functions_appinsights.py could be replaced with opentelemetry spans and tags to simplify and make the logging more robust.

image

### DIFF
--- a/docs/troubleshooting/troubleshooting.md
+++ b/docs/troubleshooting/troubleshooting.md
@@ -36,5 +36,5 @@ exceptions
 
 ## Flask Instrumentation Startup Error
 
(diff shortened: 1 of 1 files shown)


Bionic711:
Spoke with Ed offline. Accept as-is after he adds some documentation into a new troubleshooting folder under docs. He will also open an issue to track adding other settings into admin that make more sense rather than globally disabling it from the app_settings (easier access to app config than cosmos data).

PR #600, opened 2026-01-16
Retention excecution activity log

paullizer:
#### Bug Fixes

*   **Retention Policy Document Deletion Fix**
    *   Fixed critical bug where retention policy execution failed when attempting to delete aged documents, while conversation deletion worked correctly.
    *   **Root Cause 1**: Documents use `last_updated` field, but query was looking for `last_activity_at` (used by conversations).
    *   **Root Cause 2**: Date format mismatch - documents store `YYYY-MM-DDTHH:MM:SSZ` but query used Python's `.isoformat()` with `+00:00` suffix.
    *   **Root Cause 3**: Duplicate column in SELECT clause when `partition_field='user_id'` caused query errors.
    *   **Root Cause 4**: Activity logging called with incorrect `deletion_reason` parameter instead of `additional_context`.
    *   **Files Modified**: `functions_retention_policy.py` (query field names, date format, SELECT clause, activity logging).
    *   (Ref: `delete_aged_documents()`, retention policy execution, Cosmos DB queries)

### DIFF
--- a/application/single_app/functions_retention_policy.py
+++ b/application/single_app/functions_retention_policy.py
@@ -18,6 +18,7 @@ from functions_documents import delete_document, delete_document_chunks
 from functions_activity_logging import log_conversation_deletion, log_conversation_archival
 from functions_notifications import create_notification, create_group_notification, create_public_workspace_notification
 from functions_debug import debug_print
+from functions_appinsights import log_event
 from datetime import datetime, timezone, timedelta
 
 
@@ -36,6 +37,7 @@ def get_all_user_settings():
         ))
         return users
     except Exception as e:
+        log_event("get_all_user_settings_error", {"error": str(e)})
         debug_print(f"Error fetching all user settings: {e}")
         return []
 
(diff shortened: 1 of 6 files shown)


application/single_app/functions_retention_policy.py
```diff
@@ -593,27 +605,37 @@ def delete_aged_documents(retention_days, workspace_type='personal', user_id=Non
...
+    try:
+        aged_documents = list(container.query_items(
+            query=query,
+            parameters=parameters,
+            enable_cross_partition_query=True
+        ))
+        debug_print(f"Found {len(aged_documents)} aged documents for {workspace_type} workspace")
+    except Exception as query_error:
+        debug_print(f"Error querying aged documents for {workspace_type} (partition_value={partition_value}): {query_error}")
+        return {'count': 0, 'details': []}
```
Bionic711, line 656: What about non-query_error exceptions? I wonder whether one will terminate the retention policy thread.

paullizer: good catch, I've updated the logic to move the variable extraction inside the try blocks to make the code more resilient, to ensure one malformed item doesn't prevent processing of subsequent items.

added log_event to exceptions too

PR #608, opened 2026-01-16
Bug fix - fixed group member management errors with selecting new users, removing them, and updating the roles

Xeelee33:
The functions in manage_groups.js are no longer global functions in the new version, so the actions fail in the manage group screen.

Trying to select a user after search results in this error where the fields do not populate (note, this works properly in public workspaces because those functions are still global functions).
image

Additionally, the Remove button does nothing when clicked, and attempting to update a member role also throws an error. The request approval/rejection workflow also needed to be fixed.

I updated the group member select code to more closely match the code in manage_public_workspace.js for consistency and moved the inline onclick handlers for the select-user-btn, remove-member-btn, change-role-btn, approve-request-btn, and reject-request-btn to event delegation blocks in the $(document).ready function.

After pushing this change, the updated buttons and functions work as expected.

### DIFF
--- a/application/single_app/static/js/group/manage_group.js
+++ b/application/single_app/static/js/group/manage_group.js
@@ -59,6 +59,39 @@ $(document).ready(function () {
     }
   });
 
+  // Add event delegation for select user button in search results
+  $(document).on("click", ".select-user-btn", function () {
+    const id = $(this).data("user-id");
+    const name = $(this).data("user-name");
+    const email = $(this).data("user-email");
+    selectUserForAdd(id, name, email);
+  });
+
+  // Add event delegation for remove member button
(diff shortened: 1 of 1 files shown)


application/single_app/static/js/group/manage_group.js
```diff
@@ -522,66 +555,44 @@ function rejectRequest(requestId) {
   });
 }
 
+// Search users for manual add
 function searchUsers() {
   const term = $("#userSearchTerm").val().trim();
   if (!term) {
-    alert("Please enter a search term.");
+    alert("Enter a name or email to search.");
```
Bionic711, line 562: @Xeelee33 I approved the PR, but please update this away from alert() and into a toast or inline error/banner/etc.

PR #609, opened 2026-01-17
v0.235.025

paullizer:
#### Bug Fixes

*   **Retention Policy Document Deletion Fix**
    *   Fixed critical bug where retention policy execution failed when attempting to delete aged documents, while conversation deletion worked correctly.
    *   **Root Cause 1**: Documents use `last_updated` field, but query was looking for `last_activity_at` (used by conversations).
    *   **Root Cause 2**: Date format mismatch - documents store `YYYY-MM-DDTHH:MM:SSZ` but query used Python's `.isoformat()` with `+00:00` suffix.
    *   **Root Cause 3**: Duplicate column in SELECT clause when `partition_field='user_id'` caused query errors.
    *   **Root Cause 4**: Activity logging called with incorrect `deletion_reason` parameter instead of `additional_context`.
    *   **Files Modified**: `functions_retention_policy.py` (query field names, date format, SELECT clause, activity logging).
    *   (Ref: `delete_aged_documents()`, retention policy execution, Cosmos DB queries)

*   **Retention Policy Scheduler Fix**
    *   Fixed automated retention policy scheduler not executing at the scheduled time.
    *   **Root Cause 1**: Hour-matching approach was unreliable - only ran if check happened exactly during the execution hour (e.g., 2 AM), but 1-hour sleep intervals could miss the entire window.
    *   **Root Cause 2**: Check interval too long (1 hour) meant poor responsiveness and high probability of missing scheduled time.
    *   **Root Cause 3**: Code ignored the stored `retention_policy_next_run` timestamp, instead relying solely on hour matching.
    *   **Solution**: Now uses `retention_policy_next_run` timestamp for comparison, reduced check interval from 1 hour to 5 minutes, added fallback logic for missed executions.
    *   **Files Modified**: `app.py` (`check_retention_policy()` background task).
    *   (Ref: retention policy scheduler, background task, scheduled execution)


paullizer:
Testing out the Dev > Staging > Main workflow

Its missing checks and deploy code but I wanted to set up the flow, then work on what we want to add to staging.

application/single_app/app.py
```diff
@@ -326,29 +346,30 @@ def check_retention_policy():
...
+                                 f"{results['personal']['conversations']} personal conversations, "
+                                 f"{results['personal']['documents']} personal documents, "
+                                 f"{results['group']['conversations']} group conversations, "
+                                 f"{results['group']['documents']} group documents, "
+                                 f"{results['public']['conversations']} public conversations, "
+                                 f"{results['public']['documents']} public documents deleted.")
+                        else:
+                            print(f"Retention policy execution failed: {results.get('errors')}")
                 
             except Exception as e:
```
Bionic711, line 372: log_event

application/single_app/functions_retention_policy.py
```diff
@@ -36,6 +37,7 @@ def get_all_user_settings():
         ))
         return users
     except Exception as e:
+        log_event("get_all_user_settings_error", {"error": str(e)})
         debug_print(f"Error fetching all user settings: {e}")
```
Bionic711, line 41: If you are not taking advantage of any of the category or kwargs from debug print, no reason to use both. log_event has a basic built in [DEBUG] print is debugging is enabled. While we cannot use debug_print in log_event because of cyclical import issues, we could always enhance that to leverage the "extra" (custom dimensions) of log_event and output those like we do the kwargs in debug_print.

application/single_app/functions_retention_policy.py
```diff
@@ -382,6 +395,7 @@ def process_public_retention():
                     workspace_deletion_summary['conversation_details'] = conv_results['details']
                     results['conversations'] += conv_results['count']
                 except Exception as e:
+                    log_event("process_public_retention_conversations_error", {"error": str(e), "public_workspace_id": workspace_id})
```
Bionic711, line 398: Might be easier to locate these in appinsights/log analytics by setting the level to logging.error. We would probably need to import the logging lib though. It also accepts the corresponding int as well (I believe error is 40).

Bionic711 (dismissed):
@paullizer Review the couple of comments and we are good to go.

application/single_app/app.py
paullizer: added log_event across all in app.py

application/single_app/functions_retention_policy.py
paullizer: https://github.com/microsoft/simplechat/issues/612

application/single_app/functions_retention_policy.py
paullizer: https://github.com/microsoft/simplechat/issues/612

PR #623, opened 2026-01-22
Add default retention policy

paullizer:
*   **Retention Policy Defaults**
    *   Admin-configurable organization-wide default retention policies for conversations and documents across all workspace types.
    *   **Organization Defaults**: Set default retention periods (1 day to 10 years, or "Don't delete") separately for personal, group, and public workspaces.
    *   **User Choice**: Users see "Using organization default (X days)" option and can override with custom settings or revert to org default.
    *   **Conditional Display**: Default retention settings only appear in Admin Settings when the corresponding workspace type is enabled.
    *   **Force Push Feature**: Administrators can push organization defaults to all workspaces, overriding any custom retention policies users have set.
    *   **Settings Auto-Save**: Force push automatically saves pending settings changes before executing to ensure current values are pushed.
    *   **Activity Logging**: Force push actions are logged to `activity_logs` container for audit purposes with admin info, affected scopes, and results summary.
    *   **API Endpoints**: New `/api/retention-policy/defaults/<workspace_type>` (GET) and `/api/admin/retention-policy/force-push` (POST) endpoints.
    *   **Files Modified**: `functions_settings.py`, `admin_settings.html`, `route_frontend_admin_settings.py`, `route_backend_retention_policy.py`, `functions_retention_policy.py`, `functions_activity_logging.py`, `profile.html`, `control_center.html`, `workspace-manager.js`.
    *   (Ref: Default retention settings, Force Push modal, activity logging, retention policy execution)

### DIFF
--- a/application/single_app/route_backend_retention_policy.py
+++ b/application/single_app/route_backend_retention_policy.py
@@ -380,7 +380,7 @@ def register_route_backend_retention_policy(app):
             log_event(f"Force push retention defaults failed: {e}", level=logging.ERROR)
             return jsonify({
                 'success': False,
-                'error': f'Failed to push retention defaults: {str(e)}'
+                'error': f'Failed to push retention defaults'
             }), 500
     
     


application/single_app/route_backend_retention_policy.py
```diff
@@ -106,6 +107,75 @@ def update_retention_policy_settings():
...
+                    elif days == 90:
+                        return '90 days (3 months)'
+                    elif days == 180:
+                        return '180 days (6 months)'
+                    elif days == 365:
+                        return '365 days (1 year)'
+                    elif days == 730:
+                        return '730 days (2 years)'
+                    else:
+                        return f'{days} days'
```
Bionic711, line 157: It looks like we have these set here, but also in the javascript. We should consider condensing it down into a single source, creating an api call, from the UI to get this since we need it in both areas. Managing it in two places it never good.

application/single_app/route_backend_retention_policy.py
```diff
@@ -155,6 +225,162 @@ def manual_execute_retention_policy():
...
+                        user_settings = user.get('settings', {})
+                        user_settings['retention_policy'] = {
+                            'conversation_retention_days': 'default',
+                            'document_retention_days': 'default'
+                        }
+                        user['settings'] = user_settings
+                        
+                        cosmos_user_settings_container.upsert_item(user)
+                        personal_count += 1
+                    except Exception as e:
```
Bionic711, line 284: We should consider logging these with log_event, especially since they are admin events.

application/single_app/route_backend_retention_policy.py
```diff
@@ -155,6 +225,162 @@ def manual_execute_retention_policy():
...
+                        # Update workspace's retention policy to use 'default'
+                        workspace['retention_policy'] = {
+                            'conversation_retention_days': 'default',
+                            'document_retention_days': 'default'
+                        }
+                        
+                        cosmos_public_workspaces_container.upsert_item(workspace)
+                        public_count += 1
+                    except Exception as e:
+                        debug_print(f"Error updating public workspace {workspace_id}: {e}")
```
Bionic711, line 345: Same here with log_event

application/single_app/route_backend_retention_policy.py
```diff
@@ -155,6 +225,162 @@ def manual_execute_retention_policy():
...
+                'scopes': scopes,
+                'details': details
+            })
+            
+        except Exception as e:
+            debug_print(f"Error force pushing retention defaults: {e}")
+            log_event(f"Force push retention defaults failed: {e}", level=logging.ERROR)
+            return jsonify({
+                'success': False,
+                'error': f'Failed to push retention defaults: {str(e)}'
```
Bionic711: CodeQL may tag this as a high security vulnerability because it potentially exposes error information to end users. I disagree strictly because it is an admin endpoint, but we should watch for it.

Bionic711 (approved):
Nothing needed to stop the PR, but a couple notes to do in a future push.

application/single_app/route_backend_retention_policy.py
paullizer: added log_event where missing

application/single_app/route_backend_retention_policy.py
paullizer: https://github.com/microsoft/simplechat/issues/625

application/single_app/route_backend_retention_policy.py
paullizer: The admin will have access to app insights, and the console should they need more specifics. I'm removing the potential for backend data to be presented this way.

PR #722, opened 2026-02-19
MCP Server addition and modification of SimpleChat API to accomodate …

gregunger-microsoft:
Added MCP Server code with PRM auth handshake and very minimal SimpleChat API changes to accomodate auth/session.

tried to assign to Paul/Nathaniel but it is not working.
/assign @paullizer
/assign @nadoylemsft


application/external_apps/mcp/server_minimal.py
```diff
@@ -0,0 +1,1692 @@
...
+
+    if response.status_code != 200:
+        try:
+            details = response.json()
+        except Exception:
+            details = {"raw": response.text}
+        return {
+            "error": "simplechat_request_failed",
+            "status_code": response.status_code,
+            "details": details,
```
Bionic711, line 1483: We need to be careful with these return messages. Several places written in this manner and could return potentially sensitive information.

application/external_apps/mcp/server_minimal.py
Bionic711, line 1: Overall, reads like a great start and I think the features exposed are awesome. It needs some enterprise TLC (scaling via redis for session cache, appinsights/LAW logging support, etc), but a great start none the less.

application/external_apps/mcp/Dockerfile
Bionic711, line 1: Any reason this would not run in a distroless (mcr.microsoft.com/azurelinux/distroless/python:3.12) multi-step build like the core app?

application/single_app/route_external_prm.py
```diff
@@ -0,0 +1,27 @@
...
+"""Protected Resource Metadata (PRM) endpoint for SimpleChat."""
+
+from config import *
+from swagger_wrapper import swagger_route, get_auth_security
+
+
+def register_route_external_prm(app):
+    @app.route('/.well-known/oauth-protected-resource', methods=['GET'])
+    @swagger_route(security=get_auth_security())
+    def get_prm_metadata():
```
Bionic711: Does this route need to be unauthenticated (pretty sure it does)? Security risks associated with exposing the app id and tenantId to a non-authenticated endpoint? We likely will need to document this, and we need to add to code as comments that this route is explicitly unprotected by authentication (this might actually be our first one).

application/single_app/route_frontend_authentication.py
```diff
@@ -177,22 +200,28 @@ def authorized_api():
...
-            if front_door_url:
-                home_url, login_redirect_url = build_front_door_urls(front_door_url)
-                redirect_uri = login_redirect_url
-            else:
-                redirect_uri = LOGIN_REDIRECT_URL or url_for('authorized', _external=True, _scheme='https')
+        if request_redirect_uri and _is_local_redirect_uri(request_redirect_uri):
+            redirect_uri = request_redirect_uri
         else:
-            redirect_uri = url_for('authorized', _external=True, _scheme='https')
+            from functions_settings import get_settings
```
Bionic711, line 213: You are a terrible human for this.

Bionic711 (commented):
Not prod ready, but a fantastic start. We need the option to disable and enabled the mcp server from admin options, with a default to disable which includes a note that it requires a restart of the app to take effect. Scaling via Redis (at minimal for session cache and I imagine many customers will implement in app service instead of ACA as that is how most of them run it). Logging into AppInsights with OTEL.

Logging and scaling we could approve without. However, the admin settings via opt-in is required before we can get this into development. We should have a modal on the configuration page that guides the admin through the general steps to deploy the external app (platform agnostic) and, if we leave the metadata page unauthenticated, a pop-up that the admin needs to accept (and is then logged to the activity log with which admin did it) that they accepted that the tenantid and appid are exposed on an unauthenticated endpoint.

application/external_apps/mcp/server_minimal.py
gregunger-microsoft: I can return less but understand that every  place you see this code (currently) is within a tool (method) in the MCP Server itself. These specific tools are connecting to SimpleChat and the response.json() which is the "details" here is any error in the response coming from SimpleChat itself. So as long as SimpleChat is not returning sensitive information (which I assume it wouldn't/shouldn't then this should never be a problem. 

SimpleChat should never be exposing sensitive information and if it does, thats a bug. But to not forward the message (for error handling purposes) to the end client obfuscates the problem. 

If any change needs to happen here, please be prescriptive in exactly the way in which the change should be implemented. Happy to do whatever is required.

application/external_apps/mcp/server_minimal.py
gregunger-microsoft: Unclear on where you see redis coming into play here with regard to scaling? (on the MCP server side) 

SimpleChat handles session. The MCP Server only has a session created during the auth handshake and then subsequently sends a Session ID as a cookie with requests to the SimpleChat API. 

The MCP Server itself can easily scale out if needs be via horizontal scaling to accommodate concurrent user load. 

But before we get into refactoring for scaling, we should harden any patterns used by integration components, including the MCP Server as a first pass before we start coding for scaling.

Again, if there is some way you could be more prescriptive in exactly what you had in mind, I'm happy to accommodate. 

Same thing with logging here. If there is some logging pattern you want implement, let's discuss.

application/external_apps/mcp/Dockerfile
gregunger-microsoft: So no issues if distroless is a requirement. But at this point in the development, it is premature for the MCP Server itself and for non-production use-cases. We could easily have multiple docker files to accomodate distroless and not distroless.

Using a distroless container image means your container includes only your application and its runtime — no shell, no package manager, no debugging tools.

That sounds ideal, especially prematurely when adding new functionality/components to a system that may require higher visibility/observability in a given system.

Here are the main reasons why using a distroless image can be a bad idea:
#1 Debugging Becomes Painful
Distroless images don’t include:
`bash` / `sh`
`curl`
`ps`
`netstat`
`apt` / `yum`
- i.e. Most standard Linux tools

If something goes wrong in production, you can’t exec into the container and inspect it. Because there is no shell.

This makes incident response slower and more complex. You need:
- Sidecar debug containers
- Ephemeral debug images
- Rebuilds just to inspect behavior

That’s fine for mature software. Not so great for fast-moving agile software development.

#2 Local Development Becomes Friction-Filled
Distroless is great for production minimalism.
It’s terrible for:
- Experimentation
- Rapid troubleshooting
- Investigating weird runtime edge cases

Developers often end up maintaining:
- One Dockerfile for dev
- One for prod

Now you have divergence risk.

#3 Observability Tooling Limitations
Many debugging or monitoring tools assume basic OS utilities exist.
If you rely on:
- Runtime profilers
- On-the-fly tracing
- Crash dump analysis
- Temporary diagnostics
You may discover the container simply doesn’t support what you need.

#4 Operational Complexity Increases
Distroless pushes complexity outward.
Instead of “Let’s exec in and look.”
You now need:
- Debug sidecars
- Separate debug images
- Stronger logging discipline
- More complete telemetry
If your observability is weak, distroless will expose it brutally.

#5 Compatibility Surprises
Some apps assume:
- Timezone data exists
- CA certificates are present
- Locale data is available
- `/bin/sh` exists for subprocesses
Distroless images may omit or minimally include these.
You discover issues at runtime.

#6 Security Gains May Be Marginal
Distroless is often sold as “more secure.”
But:
- The real attack surface is usually your application.
- Most container exploits rely on runtime vulnerabilities, not `bash`.
- A properly hardened minimal distro (e.g., Alpine or slim Debian) may provide nearly identical risk reduction.
Security benefit is real — but sometimes overstated.

#7 Not Ideal for Dynamic Environments
If your application:
- Installs plugins dynamically
- Runs scripts
- Uses subprocess-heavy workflows
- Executes user-provided commands
Distroless can break those assumptions.

application/single_app/route_external_prm.py
gregunger-microsoft: No — a PRM authentication file does not need to be unauthenticated for MCP server.

In a typical MCP (Model Context Protocol) server–client handshake, the sequence looks like:
1 Client connects to server
2 Client presents credentials (token, key, cert, etc.)
3 Server validates credentials
4 Secure session is established
The authentication material (like a PRM file containing keys or metadata) is used during validation, not before it.
So the file does not need to be publicly accessible or unauthenticated.

application/single_app/route_frontend_authentication.py
gregunger-microsoft: I need more information here. I don't understand the critique. Is it the logic or is it doing an import statement inline? And you would prefer imports happening at the top of the file? Please be more prescriptive.

application/external_apps/mcp/server_minimal.py
Bionic711: I thought I saw where the MCP server is maintaining its own session state, if I was mistaken, then scaling is already built in by being stateless.

application/external_apps/mcp/Dockerfile
Bionic711: It will be a requirement for production (a ton of customer have requested it because of the amount of CVEs that are eliminated), but your points are valid for rapid iteration/development. That said, for rapid iteration, you should also be able to run it locally/dev container that has obserability/tools. We also prefer to rely on high and strong logging discipline.

application/single_app/route_frontend_authentication.py
Bionic711: I was just joking about the from/import statement in the middle of the code without obvious reasons (such as preventing circular references elsewhere in code, etc). @paullizer can attest these bother me.

Bionic711:
Closing this out as we have Inbound MCP in SimpleChat now. Thanks for the guidance on this!

PR #727, opened 2026-02-22
Adding agents directory to include instructions and file on how to cr…

aaronba:
Create declarative agents with OAuth settings


application/external_apps/agents/env/.env.dev
Bionic711, line 1: @aaronba is this your actual config file? If so, you need to purge it from git commit history first. If not, please replace the secrets and IDs with obvious place holder.

PR #734, opened 2026-02-24
Added ability to export conversations & set retention policies

eldong:
This PR also includes the feature to allow users to set the group and workspace retention policies


docs/explanation/features/CONVERSATION_EXPORT.md
paullizer, line 1: this is great! please also update the release_notes.md

eldong:
I updated the release notes to include the Export Conversations and Retention policy UI features.

PR #782, opened 2026-03-09
Tabular, Thoughts, PDF Export, SQL plugin updates, and bug fixes

paullizer:
#### New Features

*   **Persistent Conversation Summaries**
    *   Summaries generated during conversation export are now saved to the conversation document in Cosmos DB for future reuse.
    *   Cached summaries include `message_time_start` and `message_time_end` — when a conversation has new messages beyond the cached range, a fresh summary is generated automatically.
    *   The conversation details modal now shows a **Summary** card at the top. If a summary exists it displays the content, generation date, and model used. If no summary exists a **Generate Summary** button with model selector lets users create one on demand.
    *   A **Regenerate** button is available on existing summaries to force a refresh with the currently selected model.
    *   New `POST /api/conversations/<id>/summary` endpoint accepts an optional `model_deployment` and returns the generated summary.
    *   The `GET /api/conversations/<id>/metadata` response now includes a `summary` field.
    *   Extracted `generate_conversation_summary()` as a shared helper used by both the export pipeline and the new API endpoint.
    *   (Ref: `route_backend_conversation_export.py`, `route_backend_conversations.py`, `chat-conversation-details.js`, `functions_conversation_metadata.py`)

*   **PDF Conversation Export**
    *   Added PDF as a third export format option alongside JSON and Markdown, giving users a print-ready, visually styled conversation archive.
    *   PDF output renders chat messages with colored bubbles that mirror the live chat UI: blue for user messages, gray for assistant messages, green for file messages, and amber for system messages.
    *   Message content is converted from Markdown to HTML for rich formatting (bold, italic, code blocks, lists, tables) inside the PDF.
    *   Full appendix structure is included (metadata, message details, references, processing thoughts, supplemental messages), matching the Markdown export layout.
    *   Rendering uses PyMuPDF's Story API on US Letter paper with 0.5-inch margins and automatic multi-page overflow.
    *   Works with both single-file and ZIP packaging; intro summaries are supported in PDF as well.
    *   Frontend format step updated to a 3-column card grid with a new PDF card using the `bi-filetype-pdf` icon.
    *   (Ref: `route_backend_conversation_export.py`, `chat-export.js`, PyMuPDF Story API, conversation export workflow)

*   **Conversation Export Intro Summaries**
    *   Added an optional AI-generated intro summary step to the conversation export workflow, so each exported chat can begin with a short abstract before the full transcript.
    *   Summary model selection now reuses the same model list shown in the chat composer, keeping the export flow aligned with the main chat experience.
    *   Works for both JSON and Markdown exports, including ZIP exports where each conversation keeps its own summary metadata.
    *   (Ref: `route_backend_conversation_export.py`, `chat-export.js`, conversation export workflow)

*   **Agent & Action User Tracking (created_by / modified_by)**
    *   All agent and action documents (personal, group, and global) now include `created_by`, `created_at`, `modified_by`, and `modified_at` fields that track which user created or last modified the entity.
    *   On updates, the original `created_by` and `created_at` values are preserved while `modified_by` and `modified_at` are refreshed with the current user and timestamp.
    *   New optional `user_id` parameter added to `save_group_agent`, `save_global_agent`, `save_group_action`, and `save_global_action` for caller-supplied user tracking (backward-compatible, defaults to `None`).
    *   (Ref: `functions_personal_agents.py`, `functions_group_agents.py`, `functions_global_agents.py`, `functions_personal_actions.py`, `functions_group_actions.py`, `functions_global_actions.py`)

*   **Activity Logging for Agent & Action CRUD Operations**
    *   Every create, update, and delete operation on agents and actions now generates an activity log record in the `activity_logs` Cosmos DB container and Application Insights.
    *   Six new logging functions: `log_agent_creation`, `log_agent_update`, `log_agent_deletion`, `log_action_creation`, `log_action_update`, `log_action_deletion`.
    *   Activity records include: `user_id`, `activity_type`, `entity_type` (agent/action), `operation` (create/update/delete), `workspace_type` (personal/group/global), and `workspace_context` (group_id when applicable).
    *   Logging is fire-and-forget — failures never break the CRUD operation.
    *   All personal, group, and admin routes for both agents and actions are wired up.
    *   (Ref: `functions_activity_logging.py`, `route_backend_agents.py`, `route_backend_plugins.py`)

*   **Tabular Data Analysis — SK Mini-Agent for Normal Chat**
    *   Tabular files (CSV, XLSX, XLS, XLSM) detected in search results now trigger a lightweight Semantic Kernel mini-agent that pre-computes data analysis before the main LLM response. This brings the same analytical depth previously only available in full agent mode to every normal chat conversation.
    *   **Automatic Detection**: When AI Search results include tabular files from any workspace (personal, group, or public) or chat-uploaded documents, the system automatically identifies them via the `TABULAR_EXTENSIONS` configuration and routes the query through the SK mini-agent pipeline.
    *   **Unified Workspace and Chat Handling**: Tabular files are processed identically regardless of their storage location. The plugin resolves blob paths across all four container types (`user-documents`, `group-documents`, `public-documents`, `personal-chat`) with automatic fallback resolution if the primary source lookup fails. A user asking about an Excel file in their personal workspace gets the same analytical treatment as one asking about a CSV uploaded directly to a chat.
    *   **Six Data Analysis Functions**: The `TabularProcessingPlugin` exposes `describe_tabular_file`, `aggregate_column` (sum, mean, count, min, max, median, std, nunique, value_counts), `filter_rows` (==, !=, >, <, >=, <=, contains, startswith, endswith), `query_tabular_data` (pandas query syntax), `group_by_aggregate`, and `list_tabular_files` — all registered as Semantic Kernel functions that the mini-agent orchestrates autonomously.
    *   **Pre-Computed Results Injected as Context**: The mini-agent's computed analysis (exact numerical results, aggregations, filtered data) is injected into the main LLM's system context so it can present accurate, citation-backed answers without hallucinating numbers.
    *   **Graceful Degradation**: If the mini-agent analysis fails for any reason, the system falls back to instructing the main LLM to use the tabular processing plugin functions directly, preserving full functionality.
    *   **Non-Streaming and Streaming Support**: Both chat modes are supported. The mini-agent runs synchronously before the main LLM call in both paths.
    *   **Requires Enhanced Citations**: The tabular processing plugin depends on the blob storage client initialized by the enhanced citations system. The `enable_enhanced_citations` admin setting must be enabled for tabular data analysis to activate.
    *   (Ref: `run_tabular_sk_analysis()`, `TabularProcessingPlugin`, `collect_tabular_sk_citations()`, `TABULAR_EXTENSIONS`)

*   **Tabular Tool Execution Citations**
    *   Every tool call made by the SK mini-agent during tabular analysis is captured and surfaced as an agent citation, providing full transparency into the data analysis pipeline.
    *   **Automatic Capture**: The existing `@plugin_function_logger` decorator on all `TabularProcessingPlugin` functions records each invocation including function name, input parameters, returned results, execution duration, and success/failure status.
    *   **Citation Format**: Tool execution citations appear in the same "Agent Tool Execution" modal used by full agent mode, showing `tool_name` (e.g., `TabularProcessingPlugin.aggregate_column`), `function_arguments` (the exact parameters passed), and `function_result` (the computed data returned).
    *   **End-to-End Auditability**: Users can verify exactly which aggregations, filters, or queries were run against their data, what parameters were used, and what raw results were returned — before the LLM summarized them into the final response.
    *   (Ref: `collect_tabular_sk_citations()`, `plugin_invocation_logger.py`)

*   **SK Mini-Agent Performance Optimization**
    *   Reduced typical tabular analysis time from ~74 seconds to an estimated ~30-33 seconds (55-60% reduction) through three complementary optimizations.
    *   **DataFrame Caching**: Per-request in-memory cache eliminates redundant blob downloads. Previously, each of the ~8 tool calls in a typical analysis downloaded and parsed the same file independently. Now the file is downloaded once and subsequent calls read from cache. Cache is automatically scoped to the request (new plugin instance per analysis) and garbage-collected afterward.
    *   **Pre-Dispatch Schema Injection**: File schemas (columns, data types, row counts, and a 3-row preview) are pre-loaded and injected into the SK mini-agent's system prompt before execution begins. This eliminates 2 LLM round-trips that were previously spent on file discovery (`list_tabular_files`) and schema inspection (`describe_tabular_file`), allowing the model to jump directly to analysis tool calls.
    *   **Async Plugin Functions**: All six `@kernel_function` methods converted to `async def` using `asyncio.to_thread()`. This enables Semantic Kernel's built-in `asyncio.gather()` to truly parallelize batched tool calls (e.g., 3 simultaneous `aggregate_column` calls) instead of executing them serially on the event loop.
    *   **Batching Instructions**: The system prompt now instructs the model to batch multiple independent function calls in a single response, reducing LLM round-trips further.
    *   (Ref: `_df_cache`, `asyncio.to_thread`, pre-dispatch schema injection in `run_tabular_sk_analysis()`)

*   **SQL Test Connection Button**
    *   Added a "Test Connection" button to the SQL Database Configuration section (Step 3) of the action wizard, allowing users to validate database connectivity before saving.
    *   Supports all database types: SQL Server, Azure SQL (with managed identity), PostgreSQL, MySQL, and SQLite.
    *   Shows inline success/failure alerts with a 15-second timeout cap and sanitized error messages.
    *   New backend endpoint: `POST /api/plugins/test-sql-connection`.
    *   (Ref: `route_backend_plugins.py`, `plugin_modal_stepper.js`, `_plugin_modal.html`)

#### Bug Fixes

*   **On-Demand Summary Generation — Content Normalization Fix**
    *   Fixed the `POST /api/conversations/<id>/summary` endpoint failing with an error when generating summaries from the conversation details modal.
    *   Root cause: message `content` in Cosmos DB can be a list of content parts (e.g., `[{type: "text", text: "..."}]`) rather than a plain string. The endpoint was passing the raw list as `content_text`, which either stringified incorrectly or produced empty transcript text.
    *   Now uses `_normalize_content()` to properly flatten list/dict content into plain text, matching the export pipeline's behavior.
    *   (Ref: `route_backend_conversations.py`, `_normalize_content`, `generate_conversation_summary`)

*   **Export Summary Reasoning-Model Compatibility**
    *   Fixed export intro summary generation failing or returning empty content with reasoning-series models (gpt-5, o1, o3) through a series of incremental fixes: using `developer` role instead of `system` for instruction messages, removing all `max_tokens` / `max_completion_tokens` caps so the model decides output length naturally, and adding null-safe content extraction for `None` responses.
    *   Summary now includes ALL messages (user, assistant, system, file, image analysis) for full context, with a simplified prompt producing 1-2 factual paragraphs.
    *   Added detailed debug logging showing message count, character count, model name, role, and finish reason.
    *   (Ref: `route_backend_conversation_export.py`, `_build_summary_intro`, `generate_conversation_summary`)

*   **Conversation Export Schema and Markdown Refresh**
    *   Fixed conversation exports lagging behind the live chat schema. JSON exports now include processing thoughts, normalized citations, and the raw document/web/tool citation buckets stored with assistant messages.
    *   Fixed Markdown exports being too flat and text-heavy by reorganizing them into a transcript-first layout with appendices for metadata, message details, references, thoughts, and supplemental records.
    *   Fixed exported conversations including content that no longer matched the visible chat by filtering deleted messages and inactive-thread retries, then reapplying thread-aware ordering before export.
    *   (Ref: `route_backend_conversation_export.py`, `test_conversation_export.py`, conversation export rendering)

*   **Export Tag/Classification Rendering Fix**
    *   Fixed conversation tags and classifications rendering as raw Python dicts (e.g., `{'category': 'model', 'value': 'gpt-5'}`) in both Markdown and PDF exports.
    *   Tags now display as readable `category: value` strings, with smart handling for participant names, document titles, and generic category/value pairs.
    *   (Ref: `route_backend_conversation_export.py`, `_format_tag` helper, Markdown/PDF metadata rendering)

*   **Export Summary Error Visibility**
    *   Added `debug_print` and `log_event` logging to all summary generation error paths, including the empty-response path that previously failed silently.
    *   The actual error detail is now shown in both Markdown and PDF exports when summary generation fails, replacing the generic "could not be generated" message.
    *   (Ref: `route_backend_conversation_export.py`, `_build_summary_intro`, export error rendering)

*   **Content Safety for Streaming Chat Path**
    *   Added full Azure AI Content Safety checking to the streaming (`/api/chat/stream`) SSE path, matching the existing non-streaming (`/api/chat`) implementation.
    *   Previously, only the non-streaming path performed content safety analysis; streaming conversations bypassed safety checks entirely.
    *   Implementation includes: `AnalyzeTextOptions` analysis, severity threshold checking (severity ≥ 4 blocks the message), blocklist matching, persistence of blocked messages to `cosmos_safety_container`, creation of safety-role message documents, and proper SSE event delivery of blocked status to the client.
    *   On block, the streaming generator yields the safety message and `[DONE]` event, then stops — preventing any further LLM invocation.
    *   Errors in the content safety call are caught and logged without breaking the chat flow, consistent with the non-streaming behavior.
    *   (Ref: `route_backend_chats.py`, streaming SSE generator, `AnalyzeTextOptions`, `cosmos_safety_container`)

*   **SQL Schema Plugin — Eliminate Redundant Schema Calls**
    *   Fixed agent calling `get_database_schema` twice per query even though the full schema was already injected into the agent's instructions at load time.
    *   Root cause: The `@kernel_function` descriptions in `sql_schema_plugin.py` said "ALWAYS call this function FIRST," which overrode the schema context already available in the instructions.
    *   Updated all four function descriptions (`get_database_schema`, `get_table_schema`, `get_table_list`, `get_relationships`) to use the resilient pattern: "If the database schema is already provided in your instructions, use that directly and do NOT call this function."
    *   This eliminates ~400ms+ of unnecessary database round trips per query and aligns with the same pattern already used in `sql_query_plugin.py`.
    *   (Ref: `sql_schema_plugin.py`, `@kernel_function` descriptions, schema injection)

*   **SQL Schema Plugin — Empty Tables from INFORMATION_SCHEMA**
    *   Fixed `get_database_schema` returning `'tables': {}` (empty) despite the database having tables, while relationships were returned correctly.
    *   Root cause: SQL Server table/column enumeration used `INFORMATION_SCHEMA.TABLES` and `INFORMATION_SCHEMA.COLUMNS` views, which returned empty results in the Azure SQL environment. Meanwhile, the relationships query used `sys.foreign_keys`/`sys.tables`/`sys.columns` catalog views which worked perfectly.
    *   Migrated all SQL Server schema queries to use `sys.*` catalog views consistently: `sys.tables`/`sys.schemas` for table enumeration, `sys.columns` with `TYPE_NAME()` for column details, and `sys.indexes`/`sys.index_columns` for primary key detection.
    *   Fixed `pyodbc.Row` handling throughout the plugin — removed all `isinstance(table, tuple)` checks that could fail with pyodbc Row objects, replaced with robust try/except indexing.
    *   This enables the full schema (tables, columns, types, PKs, FKs) to be injected into agent instructions, allowing agents to construct complex multi-table JOINs for analytical queries.
    *   (Ref: `sql_schema_plugin.py`, `sys.tables`, `sys.columns`, `sys.indexes`, pyodbc.Row handling)

*   **SQL Query Plugin — Auto-Create Companion Schema Plugin**
    *   Fixed the remaining issue where SQL-connected agents still asked for clarification instead of querying the database, even after description improvements.
    *   Root cause: Agents configured with only a `sql_query` action never had a `SQLSchemaPlugin` loaded in the kernel. The descriptions demanded calling `get_database_schema` — a function that didn't exist — creating an impossible dependency that caused the LLM to ask for clarification.
    *   `LoggedPluginLoader` now automatically creates a companion `SQLSchemaPlugin` whenever a `SQLQueryPlugin` is loaded, using the same connection details. This ensures schema discovery is always available.
    *   Updated `@kernel_function` descriptions to be resilient: "If the database schema is provided in your instructions, use it directly. Otherwise, call get_database_schema." This dual-path approach works whether schema is injected via instructions or available via plugin functions.
    *   Added fallback in `_extract_sql_schema_for_instructions()` to also detect `SQLQueryPlugin` instances and create a temporary schema extractor if no `SQLSchemaPlugin` is found.
    *   (Ref: `logged_plugin_loader.py`, `sql_query_plugin.py`, `semantic_kernel_loader.py`)

*   **SQL Query Plugin Schema Awareness**
    *   Fixed agents connected to SQL databases asking users for clarification about table/column names instead of querying the database directly.
    *   Root cause: SQL Query and SQL Schema plugin `@kernel_function` descriptions were generic with no workflow guidance, agent instructions had no database schema context, and the two plugins operated independently with no linkage.
    *   Rewrote all `@kernel_function` descriptions in both SQL plugins to be prescriptive workflow guides (modeled after the working LogAnalyticsPlugin), explicitly instructing the LLM to discover schema first before generating queries.
    *   Added auto-injection of database schema into agent instructions at load time — when SQL Schema plugins are detected, the full schema (tables, columns, types, relationships) is fetched and appended to the agent's system prompt.
    *   Added new `query_database(question, query)` convenience function to `SQLQueryPlugin` for intent-aligned tool calling.
    *   Enabled the SQL-specific plugin creation path in `logged_plugin_loader.py` (was previously commented out).
    *   (Ref: `sql_query_plugin.py`, `sql_schema_plugin.py`, `semantic_kernel_loader.py`, `logged_plugin_loader.py`)

*   **Chat-Uploaded Tabular Files Now Trigger SK Mini-Agent in Model-Only Mode**
    *   Fixed an issue where tabular files (CSV, XLSX, XLS, XLSM) uploaded directly to a chat conversation were not analyzed by the SK mini-agent when no agent was selected. The model would describe what analysis it would perform instead of returning actual computed results.
    *   **Root Cause**: The mini SK agent only triggered from search results, but chat-uploaded files are stored in blob storage and not indexed in Azure AI Search. Additionally, the streaming path completely ignored `file` role messages in conversation history.
    *   **Fix**: Both streaming and non-streaming chat paths now detect chat-uploaded tabular files during conversation history building and trigger `run_tabular_sk_analysis(source_hint="chat")` to pre-compute results. The streaming path also now properly handles `file` role messages (tabular and non-tabular) matching the non-streaming path's behavior.
    *   (Ref: `route_backend_chats.py`, `run_tabular_sk_analysis()`, `collect_tabular_sk_citations()`)

*   **Group SQL Action/Plugin Save Failure**
    *   Fixed group SQL actions (sql_query and sql_schema types) failing to save correctly due to missing endpoint placeholder. Group routes now apply the same `sql://sql_query` / `sql://sql_schema` endpoint logic as personal action routes.
    *   Fixed Step 4 (Advanced) dynamic fields overwriting Step 3 (Configuration) SQL values with empty strings during form data collection. SQL types now skip the dynamic field merge entirely since Step 3 already provides all necessary configuration.
    *   Fixed auth type definition schemas (`sql_query.definition.json`, `sql_schema.definition.json`) only allowing `connection_string` auth type, blocking `user`, `identity`, and `servicePrincipal` types that the UI and runtime support.
    *   Fixed `__Secret` key suffix mismatch in additional settings schemas where `connection_string__Secret` and `password__Secret` didn't match the runtime's expected `connection_string` and `password` field names. Also removed duplicate `azuresql` enum value.
    *   (Ref: `route_backend_plugins.py`, `plugin_modal_stepper.js`, `sql_query.definition.json`, `sql_schema.definition.json`, `sql_query_plugin.additional_settings.schema.json`, `sql_schema_plugin.additional_settings.schema.json`)

#### User Interface Enhancements

*   **Agent Responded Thought — Seconds & Total Duration**
    *   The "responded" thought now shows time in **seconds** instead of milliseconds, and clarifies it is the total time from the initial user message (e.g., `'gpt-5-nano' responded (16.3s from initial message)`).
    *   A `request_start_time` is now captured at the top of both the non-streaming and streaming chat handlers, so the duration reflects the full request lifecycle — including content safety, hybrid search, and agent invocation — not just the model response time.
    *   Applies to all three agent paths: local SK agents (non-streaming), Azure AI Foundry agents, and streaming SK agents.
    *   (Ref: `route_backend_chats.py`, `request_start_time`, agent responded thoughts)

*   **Enhanced Agent Execution Thoughts**
    *   Added detailed model-level status messages during agent execution, giving users full visibility into each stage of the AI pipeline.
    *   **Model Identification**: A new "Sending to '{deployment_name}'" thought appears immediately after "Sending to agent", showing the exact model deployment being used (e.g., `gpt-5-nano`).
    *   **Generating Response**: A "Generating response..." thought now appears before the agent begins its invocation loop, matching the existing behavior for non-agent GPT calls.
    *   **Model Responded with Duration**: A "'{deployment_name}' responded ({duration}ms)" thought appears after the agent completes, showing total wall-clock execution time.
    *   Applies to all three agent paths: local SK agents (streaming and non-streaming) and Azure AI Foundry agents.
    *   Uses the existing `generation` step type (lightning bolt icon) — no frontend changes required.
    *   (Ref: `route_backend_chats.py`, `ThoughtTracker`, agent execution pipeline)

*   **List/Grid View Toggle for Agents and Actions**
    *   Added a list/grid view toggle to all four workspace areas: personal agents, personal actions, group agents, and group actions.
    *   **Grid View**: Large cards with type icon, humanized name, truncated description, and action buttons (Chat, View, Edit, Delete as applicable).
    *   **List View**: Improved table layout with fixed column widths (28%/47%/25%), humanized display names, and truncated descriptions with hover tooltips for full text.
    *   **View Button**: New eye-icon button on every agent and action that opens a read-only detail modal with gradient-header summary cards (Basic Information, Model Configuration, Instructions for agents; Basic Information, Configuration for actions).
    *   **Name Humanization**: Display names are now automatically parsed — underscores and camelCase/PascalCase boundaries are converted to properly spaced, title-cased words (e.g., `myCustomAgent` → `My Custom Agent`).
    *   **Persistent Preference**: View mode selection (list/grid) is saved per area in localStorage and restored on page load.
    *   New shared utility module `view-utils.js` provides reusable functions for all four workspace areas.
    *   (Ref: `view-utils.js`, `workspace_agents.js`, `workspace_plugins.js`, `plugin_common.js`, `group_agents.js`, `group_plugins.js`, `workspace.html`, `group_workspaces.html`, `styles.css`)

*   **Chat with Agent Button for Group Agents**
    *   Added a "Chat" button to each group agent row, allowing users to quickly select a group agent and navigate to the chat page.
    *   (Ref: `group_agents.js`, `group_workspaces.html`)

*   **Hidden Deprecated Action Types**
    *   Deprecated action types (`sql_schema`, `ui_test`, `queue_storage`, `blob_storage`, `embedding_model`) are now hidden from the action creation wizard type selector. Existing actions of these types remain functional.
    *   (Ref: `plugin_modal_stepper.js`)

*   **Advanced Settings Collapse Toggle**
    *   Step 4 (Advanced) content is now hidden behind a collapsible toggle button ("Show Advanced Settings") instead of being displayed by default. Reduces visual noise for most users.
    *   For SQL action types, the redundant additional fields UI in Step 4 is hidden entirely since all SQL configuration is already handled in Step 3.
    *   Step 5 (Summary) no longer shows the raw additional fields JSON dump for SQL types, since that data is already shown in the SQL Database Configuration summary card.
    *   (Ref: `_plugin_modal.html`, `plugin_modal_stepper.js`)


application/single_app/static/js/chat/chat-loading-indicator.js
```diff
@@ -39,12 +39,35 @@ export function showLoadingIndicatorInChatbox() {
...
+ */
+export function updateLoadingIndicatorText(text, iconClass) {
+  const textEl = document.getElementById("loading-indicator-text");
+  if (!textEl) return;
+
+  if (iconClass) {
+    textEl.innerHTML = `<i class="bi ${iconClass} me-1"></i>${text}`;
+  } else {
+    textEl.textContent = text;
+  }
```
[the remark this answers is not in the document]
paullizer: fixed

application/single_app/route_enhanced_citations.py
```diff
@@ -183,6 +184,189 @@ def get_enhanced_citation_pdf():
...
+                AND c.id = @file_id
+            """
+            items = list(cosmos_messages_container.query_items(
+                query=query_str,
+                parameters=[
+                    {'name': '@conversation_id', 'value': conversation_id},
+                    {'name': '@file_id', 'value': file_id}
+                ],
+                partition_key=conversation_id
+            ))
```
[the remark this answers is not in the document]
paullizer: fixed

application/single_app/route_enhanced_citations.py
```diff
@@ -183,6 +184,189 @@ def get_enhanced_citation_pdf():
...
+    @user_required
+    @enabled_required("enable_enhanced_citations")
+    def get_enhanced_citation_tabular_preview():
+        """
+        Return JSON preview of a tabular file for rendering as an HTML table.
+        Reads the file into a pandas DataFrame and returns columns + rows as JSON.
+        """
+        doc_id = request.args.get("doc_id")
+        max_rows = min(int(request.args.get("max_rows", 200)), 500)
+        if not doc_id:
```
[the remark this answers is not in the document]
paullizer: fixed

application/single_app/route_enhanced_citations.py
```diff
@@ -183,6 +184,189 @@ def get_enhanced_citation_pdf():
...
+            elif ext in ('xlsx', 'xlsm'):
+                df = pandas.read_excel(io.BytesIO(data), engine='openpyxl', keep_default_na=False, dtype=str)
+            elif ext == 'xls':
+                df = pandas.read_excel(io.BytesIO(data), engine='xlrd', keep_default_na=False, dtype=str)
+            else:
+                return jsonify({"error": f"Unsupported file type: {ext}"}), 400
+
+            total_rows = len(df)
+            preview = df.head(max_rows)
+
```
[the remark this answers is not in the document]
paullizer: capped at 500 mb but admin configurable in admin settings. files larger than the limit are not displayed but are still used its just related to displaying the file. we can revisit this later, if its an issue across orgs

application/single_app/functions_global_actions.py
```diff
@@ -76,8 +77,27 @@ def save_global_action(action_data):
...
+
+        if existing_action:
+            action_data['created_by'] = existing_action.get('created_by', user_id)
+            action_data['created_at'] = existing_action.get('created_at', now)
+        else:
+            action_data['created_by'] = user_id
+            action_data['created_at'] = now
+        action_data['modified_by'] = user_id
+        action_data['modified_at'] = now
+        action_data['updated_at'] = now
```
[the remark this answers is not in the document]
paullizer: fixed

application/single_app/semantic_kernel_plugins/sql_query_plugin.py
Bionic711, line 1: Has this already been made compatible with Key Vault storage? I know when it was originally created, it was not because of all the custom work that you did on it and the UI/etc.

application/single_app/semantic_kernel_plugins/tabular_processing_plugin.py
```diff
@@ -0,0 +1,1366 @@
...
+    SUPPORTED_EXTENSIONS = {'.csv', '.xlsx', '.xls', '.xlsm'}
+    DAY_NAME_ORDER = [
+        'Monday',
+        'Tuesday',
+        'Wednesday',
+        'Thursday',
+        'Friday',
+        'Saturday',
+        'Sunday'
+    ]
```
Bionic711: We should move these to config.py as global variables

application/single_app/static/json/schemas/sql_query_plugin.additional_settings.schema.json
Bionic711, line 1: This prevents those secrets from being stored in keyvault if enabled. I recommend we change this to include migration code for the settings.

application/single_app/static/json/schemas/sql_schema_plugin.additional_settings.schema.json
Bionic711, line 1: This prevents those secrets from being stored in keyvault if enabled.

application/single_app/functions_content.py
Bionic711, line 1: Do the 429 responses not contain a wait time?

application/single_app/route_backend_chats.py
```diff
@@ -28,24 +29,811 @@
...
+    'query_tabular_data',
+    'group_by_aggregate',
+    'group_by_datetime_component',
+}
+TABULAR_THOUGHT_EXCLUDED_PARAMETER_NAMES = {
+    'user_id',
+    'conversation_id',
+    'group_id',
+    'public_workspace_id',
+}
```
Bionic711: These should probably be moved out of this and either to the plugin or config.py as they are not directly related to the class. Also, we should consider making this a method on the plugin so that anything can call and retrieve these. This way closely couples the function in the plugin with this route, which is not good design.

It is simple for implementation. We can either fix it now or add it to the backlog for later.

application/single_app/route_backend_chats.py
Bionic711, line 1: I think we should take the entire tabular process out of this file and into its to, to provide separation, reusability, and simplification. This one (technically two) routes are already complicated enough.

application/single_app/route_backend_chats.py
```diff
@@ -28,24 +29,811 @@
...
+                "tabular_processing plugin functions before answering. Never answer from "
+                "the schema preview alone. Never say that you would need to run the "
+                "analysis later — run it now.\n\n"
+                f"{retry_prefix}"
+                f"{tool_error_feedback}"
+                f"{sheet_hint_feedback}"
+                f"{missing_sheet_feedback}"
+                f"FILE SCHEMAS:\n"
+                f"{schema_context}\n\n"
+                "AVAILABLE FUNCTIONS: lookup_value, aggregate_column, filter_rows, query_tabular_data, "
```
Bionic711, line 1550: Move this to a markdown and parse it in and move the functions to another file or the plugin, and call that method.

application/single_app/route_backend_chats.py
Bionic711, line 1: This is screaming for either session-based kernel implementation or moving up and redesigning the existing SK loader to be coherent for the entire session, and adding the tabular agent to that kernel, alongside the other agents as well.

Nothing to do now but is it leading us towards our "SK always on" mode anyway.

test_output.txt
Bionic711, line 1: Do we need?

test_output2.txt
Bionic711, line 1: Do we need?

test_results_tabular.json
Bionic711, line 1: Do we need?

artifacts/faa_flight_operations_dataset.csv
Bionic711, line 1: Do we need?

Bionic711 (changes requested):
I will approve after the conflicts are solved because I do not want to review again.

application/single_app/semantic_kernel_plugins/sql_query_plugin.py
paullizer: added in support for key vault

application/single_app/semantic_kernel_plugins/tabular_processing_plugin.py
paullizer: updated to  SUPPORTED_EXTENSIONS = tuple(f'.{extension}' for extension in sorted(TABULAR_EXTENSIONS))

application/single_app/static/json/schemas/sql_query_plugin.additional_settings.schema.json
paullizer: resolved by adding in key vault support

application/single_app/static/json/schemas/sql_schema_plugin.additional_settings.schema.json
paullizer: resolved by adding in key vault support

application/single_app/functions_content.py
paullizer: it does, updated logic. The embedding retry helpers now honor wait times returned by 429 responses

application/single_app/route_backend_chats.py
paullizer: moved to tabular processing plugin

application/single_app/route_backend_chats.py
paullizer: Im going to make this update later https://github.com/microsoft/simplechat/issues/802

application/single_app/route_backend_chats.py
paullizer: Im going to make this update later https://github.com/microsoft/simplechat/issues/802

application/single_app/route_backend_chats.py
paullizer: 100%

test_output.txt
paullizer: removed

test_output2.txt
paullizer: removed

test_results_tabular.json
paullizer: removed

artifacts/faa_flight_operations_dataset.csv
paullizer: removed

application/single_app/static/js/plugin_modal_stepper.js
```diff
@@ -538,43 +555,52 @@ export class PluginModalStepper {
...
-                    console.error('Error populating dynamic additional fields:', error);
+                  } else {
+                    console.log('No additional settings schema found');
+                    additionalFieldsDiv.classList.add('d-none');
                   }
-                } else {
-                  console.log('No additional settings schema found');
+                })
+                .catch(error => {
+                  console.error(`Error fetching additional settings schema for type: ${this.selectedType} -- ${error}`);
```
Bionic711, line 593: Is instrumenting Client-side app insights a bridge too far?

application/single_app/templates/chats.html
```diff
@@ -383,27 +385,99 @@ <h5 id="current-conversation-title" class="mb-0 flex-grow-1">
...
                         </button>
 
+                        </div>
+
                         <!-- Prompt Selector -->
+                        <div class="chat-toolbar-selectors">
                         {% if settings.enable_user_workspace or settings.enable_group_workspaces %}
-                        <div id="prompt-selection-container" class="flex-shrink-0" style="display: none;">
-                            <select class="form-select" id="prompt-select"></select>
+                        <div id="prompt-selection-container" class="chat-toolbar-selector" style="display: none;">
```
Bionic711, line 349: Does the d-none class not work here?

application/single_app/templates/chats.html
```diff
@@ -383,27 +385,99 @@ <h5 id="current-conversation-title" class="mb-0 flex-grow-1">
...
+                                </div>
+                            </div>
+                            <select class="form-select d-none" id="prompt-select"></select>
                         </div>
                         {% endif %}
 
 
                         <!-- Agent selector -->
-                        <div id="agent-select-container" style="display: none;">
+                        <div id="agent-select-container" class="chat-toolbar-selector" style="display: none;">
```
Bionic711, line 376: d-none here?

Bionic711 (dismissed):

application/single_app/templates/chats.html
Bionic711: https://github.com/microsoft/simplechat/issues/803

application/single_app/templates/chats.html
Bionic711: https://github.com/microsoft/simplechat/issues/803

application/single_app/static/js/chat/chat-enhanced-citations.js
```diff
@@ -18,11 +18,13 @@ export function getFileType(fileName) {
     const imageExtensions = ['jpg', 'jpeg', 'png', 'bmp', 'tiff', 'tif'];
     const videoExtensions = ['mp4', 'mov', 'avi', 'mkv', 'flv', 'webm', 'wmv', 'm4v', '3gp'];
     const audioExtensions = ['mp3', 'wav', 'ogg', 'aac', 'flac', 'm4a'];
-    
+    const tabularExtensions = ['csv', 'xlsx', 'xls', 'xlsm'];
```
Bionic711, line 21: We should consider having stuff like this as an API call to the backend and then caching client side. One throat to choke as we change, update, add, remove file types. Also keeps it consistent with other places we have done this.

application/single_app/static/js/chat/chat-streaming.js
```diff
@@ -21,7 +38,7 @@ export function initializeStreamingToggle() {
     // Load initial state from user settings
     loadUserSettings().then(settings => {
         console.log('Loaded user settings:', settings);
-        streamingEnabled = settings.streamingEnabled === true;
+        streamingEnabled = true;
```
Bionic711: If I read this correctly, we are completely removing the non-streaming option? Why keep the button around at all (or at least the function below to update the button state and visibility)?

application/single_app/app.py
```diff
@@ -667,16 +553,22 @@ def list_semantic_kernel_plugins():
...
 if __name__ == '__main__':
-    settings = get_settings(use_cosmos=True)
-    app_settings_cache.configure_app_cache(settings, get_redis_cache_infrastructure_endpoint(settings.get('redis_url', '').strip().split('.')[0]))
-    app_settings_cache.update_settings_cache(settings)
-    initialize_clients(settings)
-
     debug_mode = os.environ.get("FLASK_DEBUG", "0") == "1"
+    use_gunicorn = os.environ.get("SIMPLECHAT_USE_GUNICORN", "0").strip().lower() in ('1', 'true', 'yes', 'on')
+
+    if use_gunicorn and not debug_mode:
```
Bionic711, line 566: What if gunicorn is on and so is debug mode? We mislead the admin that enables gunicorn but also enables debug mode into believing they are running gunicorn.

Bionic711 (changes requested):
Only thing I saw that NEEDS to be addressed would be the gunicorn and debug mode on.

Bionic711 (dismissed):

Bionic711 (dismissed):

PR #812, opened 2026-03-27
Feature/aifoundryagents

Bionic711:
Adds multiendpoint support for all 3 scopes (models, as well as classic and new foundry agent support)
Fixes for a multitude of stuff
Updates core plugins to 1st class by adding plugin logger
Enhances thoughts and streaming
Improves MS graph plugin capabilities

### DIFF
--- a/application/single_app/functions_prompts.py
+++ b/application/single_app/functions_prompts.py
@@ -101,8 +101,6 @@ def _get_public_prompt_items(prompt_type, public_workspace_id):
 
 
 def _get_prompt_doc_with_container(user_id, prompt_id, prompt_type, group_id=None, public_workspace_id=None):
-    del user_id
-
     if public_workspace_id is not None:
         item = _read_prompt_from_container(
             cosmos_public_prompts_container,
--- a/application/single_app/config.py
+++ b/application/single_app/config.py
@@ -94,7 +94,7 @@ load_dotenv()
 EXECUTOR_TYPE = 'thread'
 EXECUTOR_MAX_WORKERS = 30
 SESSION_TYPE = 'filesystem'
-VERSION = "0.239.205"
+VERSION = "0.239.206"
 
 SECRET_KEY = os.getenv('SECRET_KEY', 'dev-secret-key-change-in-production')
 
--- a/docs/explanation/fixes/PERSONAL_PROMPT_HELPER_USER_ID_FIX.md
+++ b/docs/explanation/fixes/PERSONAL_PROMPT_HELPER_USER_ID_FIX.md
@@ -1,0 +1,59 @@
+# Personal Prompt Helper User Id Fix
+
+Fixed/Implemented in version: **0.239.206**
+
+## Issue Description
+
+Personal prompt retrieval, update, and delete operations started failing after the shared prompt helper was refactored to support more workspace scopes. The create and list flows still worked, which made the regression easy to miss until users tried to reopen or edit an existing personal prompt.
+
+## Root Cause Analysis
+
+The shared helper in `functions_prompts.py` deleted the `user_id` argument at the top of `_get_prompt_doc_with_container()`. The public and group branches did not need that variable, but the personal-scope fallback still used `user_id` to constrain the lookup. Once execution reached that branch, Python raised an `UnboundLocalError` instead of resolving the user-scoped prompt.
+
+## Technical Details
+
+Files modified:
+
+- `application/single_app/functions_prompts.py`
+- `application/single_app/config.py`
+- `functional_tests/test_personal_prompt_helper_user_id_fix.py`
+
+Code changes summary:
+
+- Removed the local deletion of `user_id` from `_get_prompt_doc_with_container()`.
+- Added a regression test that executes personal prompt create, get, update, and delete behavior against stub Cosmos containers.
+- Bumped `config.py` to version `0.239.206`.
+
+## Testing Approach
+
+Functional validation was added in `functional_tests/test_personal_prompt_helper_user_id_fix.py`.
+
+The test verifies:
+
+- personal prompt create, get, update, and delete flows work through the shared helper without raising runtime errors
+- prompt ownership remains scoped to the caller user ID
+- non-owner update and delete attempts continue to fail safely
+- the application version is updated to `0.239.206`
+
+## Impact Analysis
+
+Before:
+
+- personal prompt GET, PATCH, and DELETE flows could return 500 errors
+- prompt ownership checks in the shared helper never reached their intended personal-scope branch
+
+After:
+
+- personal prompt CRUD lookups resolve correctly through the shared helper
+- owner-only behavior is preserved for personal prompts
+- the regression is covered by a focused helper-level functional test
+
+## Validation
+
+Test results:
+
+- `functional_tests/test_personal_prompt_helper_user_id_fix.py`
+
+User experience improvements:
+
+- users can reopen, edit, and delete their saved personal prompts again without server errors
--- a/functional_tests/test_personal_prompt_helper_user_id_fix.py
+++ b/functional_tests/test_personal_prompt_helper_user_id_fix.py
@@ -1,0 +1,190 @@
+#!/usr/bin/env python3
+# test_personal_prompt_helper_user_id_fix.py
+"""
+Functional test for personal prompt helper user ID handling.
+Version: 0.239.206
+Implemented in: 0.239.206
+
+This test ensures that personal prompt get, update, and delete flows keep
+using the caller user_id after the shared prompt helper refactor.
+"""
+
+from datetime import datetime, timezone
+from pathlib import Path
+import sys
+import types
+import uuid
+
+
+ROOT_DIR = Path(__file__).resolve().parents[1]
+PROMPTS_FILE = ROOT_DIR / "application" / "single_app" / "functions_prompts.py"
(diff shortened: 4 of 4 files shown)


application/single_app/route_backend_models.py
paullizer, line 1: user/group model fetch and test endpoints bypass the custom-endpoint feature gates

44-50: resolve_request_endpoint_payload() merges raw request payload even when no persisted endpoint_id exists.
208-286: both handle_fetch_model_list() and handle_test_model_connection() consume that resolved payload.
823-836: /api/user/models/fetch and /api/user/models/test-model only require login/user auth.
839-866: /api/group/models/fetch and /api/group/models/test-model only require group workspaces, not allow_group_custom_endpoints.

Why this matters:

These routes will accept a caller-supplied endpoint/auth payload when endpoint_id is omitted.
That means the backend can still be used to probe arbitrary AOAI/Foundry endpoints even when custom endpoints are disabled.
For the user routes this is available to any authenticated user. For the group routes it is available even if group custom endpoints are disabled.
Impact:

Feature gating is not actually enforced on the backend.
The server becomes a credentialed network proxy for endpoint discovery and test calls outside the intended admin/workspace flows.

Suggested review comment:

Require the same enabled_required(...) guards as the save/list routes, and reject raw payloads for non-admin scopes unless they reference a persisted endpoint the caller is authorized to use.

application/single_app/static/js/chat/chat-streaming.js
paullizer, line 1: stream reattach cannot work for the first message in a brand-new conversation

application/single_app/route_backend_chats.py
application/single_app/static/js/chat/chat-streaming.js


5324-5325: stream session is started from the incoming conversation_id before the backend creates a new one.
5669-5670: the streaming path generates a new UUID later if the request started without a conversation id.
7143-7166: status/reattach endpoints only look up sessions by the final conversation id.
286-302 in chat-streaming.js: the client reattach flow requires a real conversation id.

Why this matters:

When a user starts streaming in a new conversation, there is no incoming conversation_id, so no live session is registered.
If the page reconnects during that first response, the reattach/status endpoints have nothing to find under the eventual conversation id.

Impact:

Reattach works for existing conversations but not for the first streamed reply of a new conversation, which is exactly when the connection is most fragile.

Suggested review comment:

Start/register the stream session after the conversation id is finalized, or allow the session key to be rebound once the UUID is created.

msgraph_test_output.txt
paullizer, line 1: delete please

application/single_app/route_backend_models.py
```diff
@@ -185,4 +504,371 @@ def get_image_models():
...
+                continue
+            endpoint_id = endpoint.get("id")
+            if endpoint_id and endpoint_id not in saved_endpoint_ids:
+                keyvault_model_endpoint_delete_helper(endpoint, endpoint_id, scope="user")
+
+        update_user_settings(user_id, {"personal_model_endpoints": saved_endpoints})
+        return jsonify({"success": True})
+
+
+    @app.route('/api/group/model-endpoints', methods=['GET'])
```
paullizer, line 798: proup model endpoint reads trust activeGroupOid without verifying current membership


application/single_app/route_backend_models.py
application/single_app/functions_group.py


660-675 in route_backend_models.py: /api/group/model-endpoints GET only calls require_active_group(user_id).
132-149 in functions_group.py: require_active_group() only reads the stored active group id, while assert_group_role() is the function that actually enforces membership.

Why this matters:

The read endpoint returns sanitized group model endpoints based only on the caller's stored activeGroupOid.
Unlike the write/test/list-foundry paths, it never verifies that the caller is still a member of that group.

Impact:

A stale or tampered activeGroupOid can expose group endpoint metadata to a user who no longer belongs to the group.
This is inconsistent with the stricter authorization already used by the adjacent POST/test routes.

Suggested review comment:

Apply the same assert_group_role(..., allowed_roles=("Owner", "Admin", "DocumentManager", "User")) check on the GET route before returning group endpoint metadata.

application/single_app/semantic_kernel_loader.py
```diff
@@ -233,8 +275,232 @@ def get_global_gpt():
...
+                return build_token_provider(auth_settings, provider="aoai")
+            except Exception as e:
+                log_event(
+                    f"[SK Loader] Failed to build global GPT token provider: {e}",
+                    level=logging.ERROR,
+                    exceptionTraceback=True,
+                )
+        return None
+
+    def resolve_multi_endpoint_agent_config():
```
paullizer, line 351: group agents resolve model endpoints from the current active group, not from the agent's own group

application/single_app/semantic_kernel_loader.py
application/single_app/functions_group_agents.py
application/single_app/route_backend_agents.py

359-368 and 417-428 in semantic_kernel_loader.py: when is_group_agent is true, endpoint lookup uses require_active_group(get_current_user_id()).
66-71 in functions_group_agents.py: group agents are persisted with their own group_id.
60-73 and 90-99 in route_backend_agents.py: selected-agent payload matching also preserves group_id for group agents.

Why this matters:

The selected group agent already carries its own group identity, but model endpoint resolution ignores that and re-derives the group from the user's current active group.
If the user changes active group after selecting an agent, or opens a conversation tied to a different group context, the loader can resolve the wrong endpoint set.

Impact:

Group agents can fail to load their configured endpoint/model.
Worse, a group agent may resolve against another active group's endpoints if endpoint ids collide or if the intended endpoint is missing in the current group.

Suggested review comment:

Resolve group-scoped model endpoints using the agent's persisted group_id (or the conversation's group id), not the mutable active-group setting.

application/single_app/semantic_kernel_loader.py
```diff
@@ -119,20 +161,20 @@ def resolve_agent_config(agent, settings):
     user_apim_enabled = agent.get("enable_agent_gpt_apim") in [True, 1, "true", "True"]
     global_apim_enabled = settings.get("enable_gpt_apim", False)
     per_user_enabled = settings.get('per_user_semantic_kernel', False)
-    allow_user_custom_agent_endpoints = settings.get('allow_user_custom_agent_endpoints', False)
-    allow_group_custom_agent_endpoints = settings.get('allow_group_custom_agent_endpoints', False)
+    allow_user_custom_endpoints = settings.get('allow_user_custom_endpoints', False) or settings.get('allow_user_custom_agent_endpoints', False)
```
paullizer, line 134: loader feature-flag typo makes group custom endpoint behavior depend on the user custom-agent flag

application/single_app/semantic_kernel_loader.py

164-165 in semantic_kernel_loader.py:
allow_user_custom_endpoints = settings.get('allow_user_custom_endpoints', False) or settings.get('allow_user_custom_agent_endpoints', False)
allow_group_custom_endpoints = settings.get('allow_group_custom_endpoints', False) or settings.get('allow_user_custom_agent_endpoints', False)

Why this matters:

The second line appears to reference allow_user_custom_agent_endpoints instead of the group equivalent.
That means enabling user custom-agent endpoints can inadvertently enable group custom-endpoint resolution logic inside the loader.

Impact:

Group-agent routing behavior becomes dependent on the wrong admin toggle.
This creates inconsistent runtime behavior between UI settings and backend agent resolution.

Suggested review comment:

Replace the second fallback with the intended group-scoped flag so group endpoint resolution only activates under group endpoint settings.

msgraph_test_output.txt
Bionic711: deleted

application/single_app/route_backend_models.py
Bionic711: Added feature gates. Implemented in the resolver that if an endpoint ID is provided, it must resolve to a persisted scope preventing probing, but allowing fetch/get of new endpoints to persist and enable user/group modal functionality.

application/single_app/route_backend_models.py
Bionic711: Added assert to missing routes. (there were actually 2 I found).

application/single_app/semantic_kernel_loader.py
Bionic711: Resolved

application/single_app/semantic_kernel_loader.py
Bionic711: Resolved

application/single_app/static/js/chat/chat-streaming.js
Bionic711: Resolved

application/single_app/functions_prompts.py
```diff
@@ -21,6 +21,134 @@ def get_pagination_params(args):
...
+    combined_items = {}
+    for item in primary_items + legacy_items:
+        prompt_id = item.get('id')
+        if prompt_id and prompt_id not in combined_items:
+            combined_items[prompt_id] = item
+
+    return _sort_prompt_items(list(combined_items.values()))
+
+
+def _get_prompt_doc_with_container(user_id, prompt_id, prompt_type, group_id=None, public_workspace_id=None):
```
paullizer, line 103: Personal prompt GET, PATCH, and DELETE now hard-fail on the personal scope. The new helper in functions_prompts.py:103 deletes user_id at functions_prompts.py:104, then tries to reuse it at functions_prompts.py:144. The personal prompt routes at route_backend_prompts.py:65, route_backend_prompts.py:85, and route_backend_prompts.py:119 all hit that path, so viewing, editing, or deleting an existing personal prompt will return 500.

paullizer: removed the stray user_id deletion

PR #917, opened 2026-06-18
Add configurable DLP controls with optional Presidio endpoint support

zarg-dev:
This PR adds a configurable DLP layer for web-search egress and upload ingestion. The built-in path is regex-based with narrow defaults, and admins can tune or expand the rules. I also added optional integration with an external Presidio Analyzer-compatible endpoint for customers who want richer detectors without adding Presidio packages and models to the SimpleChat app image.

## What changed

- Added DLP evaluation before web-search grounding.
- Added DLP evaluation during upload ingestion before embeddings/search indexing.
- Added admin-configurable regex rules with defaults for U.S. SSNs and Luhn-valid credit cards.
- Added keyword-proximity confidence shaping for regex rules.
- Added monitor, redact, and block modes.
- Added optional external Presidio Analyzer endpoint support.
- Kept Presidio as a separate service integration instead of embedding it into SimpleChat.
- Added endpoint hardening for Presidio calls, including private-host checks, DNS rebinding protection, env-backed auth, and auth-header validation.
- Added admin settings, docs, release notes, and functional tests.

## Notes

The regex model is intentionally simple. It gives deployers a lightweight starting point, but Presidio is the better fit for customers who want broader PII detection such as emails, phone numbers, IP addresses, names, or other recognizers.

I extended the same DLP path to uploads because the use case felt similar: some customers may want to scrub or block sensitive content before it is sent to model inference, embeddings, or search indexing.

## Testing

- Ran focused DLP tests for web search, uploads, regex rules, telemetry, admin settings, and Presidio endpoint behavior.
- Ran direct Presidio functional test entrypoints.

---

## Merge-resolution changes outside the DLP feature

This branch was merged with `Development` up to `4a29f67e`. Three changes below
are not part of the DLP control plane but were required for the merged branch to
land green. They are called out separately so they can be reviewed on their own
terms.

### 1. `scripts/build_release_notes_pages.py` — relative link rewriting

`docs/explanation/release_notes.md` is the source of truth and the pages under
`docs/explanation/release-notes/` are generated from it. `test_docs_release_notes_integrity`
requires the two to agree, so adding a release-notes section means regenerating.

The generated pages sit one directory below the source, but the generator copied
release text verbatim. Relative links such as `fixes/NAME.md`, which resolve
correctly from `docs/explanation/`, therefore pointed at files that do not exist
once a section reached a generated page — 8 broken links, caught by
`test_docs_link_integrity`.

`rewrite_relative_links()` now prefixes `../` for relative targets, skipping
absolute paths, anchors, URL schemes and Liquid expressions. This is a
pre-existing generator bug that regeneration exposed rather than introduced;
the committed pages were already stale on `Development` (the index covered only
up to v0.261.003 while the source had later versions).

### 2. Documentation for the DLP capability toggles

`test_docs_app_surface_coverage` fails on any unclaimed `enable_*` settings key.
This PR adds three: `enable_dlp_control_plane`, `enable_upload_dlp` and
`enable_web_search_dlp`.

- Claimed in `docs/_data/features.yml` under a new `data-loss-prevention` entry.
- Documented on `docs/admin/security.md` under Content Safety, covering the
  monitor/redact/block modes, the regex and external Presidio engines, the
  private-host allowlist, env-var-backed API key handling, and counts-only
  telemetry.
- `docs/_data/app_surface.yml` regenerated with `scripts/build_docs_inventory.py`.

### 3. `functions_dlp_presidio.py` — urllib3 version safety

The adapter imported `urllib3.util.timeout._DEFAULT_TIMEOUT`, a private symbol
that only exists on urllib3 2.x, and referenced `urllib3.connection.NameResolutionError`,
also 2.x-only. `urllib3` is not pinned and is resolved transitively through
`requests`, and `functions_documents.py` imports this module, so on urllib3 1.x
the private import raised `ImportError` during application start-up.

The sentinel now resolves through urllib3's public `Timeout.DEFAULT_TIMEOUT`,
which is `is`-identical to the 2.x private sentinel and to
`socket._GLOBAL_DEFAULT_TIMEOUT` on 1.x. Connect-timeout behavior on the DLP
request path is unchanged: an explicit timeout is still applied and only the
"leave the socket default alone" case skips `settimeout()`. DNS failures raise
`NameResolutionError` where available and its base `NewConnectionError`
otherwise. No `urllib3` pin was added.

Covered by `functional_tests/test_dlp_presidio_urllib3_compatibility.py`, which
passes on both urllib3 1.26.20 and 2.7.0.

### Template placement note

`Development` split `templates/admin_settings.html` into `templates/admin/_panes/*.html`.
The DLP card was moved verbatim into `admin/_panes/content-safety.html`,
immediately after the Content Safety card, matching its original position
relative to its neighbours. Left inline in the old monolith it would have
rendered outside every `tab-pane`. The CSRF hidden input was reapplied to the
real `<form>` in `admin_settings.html`, and three DLP tests were repointed at
the pane.

Three PR-added `redirect(url_for('admin_settings'))` calls — two on DLP
validation-failure paths — were updated to
`url_for('frontend_admin_settings.admin_settings')` after `Development` moved
that route into a blueprint. As merged they would have raised `BuildError`.


zarg-dev:
@microsoft-github-policy-service agree company="OCC"

application/single_app/functions_authentication.py
```diff
@@ -2,9 +2,14 @@
...
 import json
+import re
 
 from config import *
-from functions_appinsights import log_event
+try:
+    from functions_appinsights import log_event
+except Exception:
+    def log_event(message, extra=None, level=None, exceptionTraceback=False):
+        return None
```
Bionic711, line 12: Was there an issue trying to import log_event? Why wrap it in the try block?

application/single_app/functions_dlp.py
```diff
@@ -0,0 +1,506 @@
+# functions_dlp.py
+
+import hashlib
+import logging
+import re
+from collections import OrderedDict
+
+from functions_dlp_presidio import analyze_with_presidio_endpoint
```
Bionic711, line 8: presidio was mentioned to be optional but here it appears to be joining the core application alongside the new dlp feature?

Bionic711:
@zarg-dev @paullizer  and I are discussing the implementation and will get back to you once we have feedback. Stay tuned!

PR #1081, opened 2026-07-27
Fix Cosmos container startup race during local Docker boot

chadpalmer:
## Summary

- Fixes a local Docker startup failure where multiple gunicorn workers could race while creating first-run Cosmos containers.
- Adds conflict recovery around Cosmos `create_container_if_not_exists` so a worker re-reads and returns the container if another worker created it first.
- Adds focused regression coverage and fix documentation.
- Updates release notes under `v0.250.062`.

## Validation

- `git diff --check upstream/Development...HEAD`
- `python -m py_compile application/single_app/config.py functional_tests/test_cosmos_container_conflict_recovery.py`
- `python functional_tests/test_cosmos_container_conflict_recovery.py`

## Notes

- Branch was merged with latest `upstream/Development`.
- No route, UI, or browser-rendering files are changed in the final PR delta.
- Full Docker startup smoke test was not run locally as part of this validation, but local Docker instance did start up successfully after the updates.

### DIFF
--- a/application/single_app/config.py
+++ b/application/single_app/config.py
@@ -96,7 +96,7 @@ load_dotenv()
 EXECUTOR_TYPE = 'thread'
 EXECUTOR_MAX_WORKERS = 30
 SESSION_TYPE = 'filesystem'
-VERSION = "0.250.057"
+VERSION = "0.250.058"
 
 SESSION_COOKIE_SAMESITE = os.getenv('SESSION_COOKIE_SAMESITE', 'Lax')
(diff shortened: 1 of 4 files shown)


chadpalmer:
@Bionic711 @paullizer This PR fixes a race condition for local docker instances where the gunicorn workers trip over each other in creating the first-run Cosmos containers.

paullizer:
I like this. Nice Chat! I also appreciated the other issues on UX and 508. Thank you!

PR #1174, opened 2026-08-07
Fix exhaustive tabular CSV runs to queue directly from source

paullizer:
Summary:
- Bypass mini-agent paging for explicit exhaustive row-level CSV export/analysis requests when exactly one authorized CSV source is selected or uploaded.
- Build a replayable full-source query descriptor directly from the authorized blob, count rows through the bounded CSV replay path, and queue the existing durable runner for 300/3,000/30,000-row cases.
- Wire the direct queue path through mixed-source, legacy workspace-search, chat-upload, and streaming branches while preserving mini-agent fallback for unsupported sources.
- Add a regression test proving direct source-backed queueing works without prior filter_rows/query_tabular_data invocations.
- Bump app version to 0.250.133.

Log analysis:
- Captured App Service logs showed zero filter_rows/query_tabular_data/generated-output routing for the failed attempts, so the durable runner was never reached. This change moves exhaustive CSV requests ahead of that fragile dependency.

Validation:
- .\\.venv\\Scripts\\python.exe -m py_compile application\\single_app\\route_backend_chats.py application\\single_app\\config.py functional_tests\\test_tabular_row_orchestration_scale.py
- .\\.venv\\Scripts\\python.exe functional_tests\\test_tabular_row_orchestration_scale.py

### DIFF
--- a/application/single_app/config.py
+++ b/application/single_app/config.py
@@ -96,7 +96,7 @@ DOTENV_LOAD_RESULT = load_simplechat_dotenv()
 EXECUTOR_TYPE = 'thread'
 EXECUTOR_MAX_WORKERS = 30
 SESSION_TYPE = 'filesystem'
-VERSION = "0.250.132"
+VERSION = "0.250.133"
 IS_DEVELOPMENT = is_development_env_enabled()
 
 SESSION_COOKIE_SAMESITE = os.getenv('SESSION_COOKIE_SAMESITE', 'Lax')
--- a/application/single_app/route_backend_chats.py
+++ b/application/single_app/route_backend_chats.py
@@ -64,6 +64,7 @@ import csv
 import io
 import inspect
 import json
+import math
 import mimetypes
 import os
 import app_settings_cache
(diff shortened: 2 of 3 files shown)


paullizer:
Follow-up pushed after live log review:
- Removed local traceback imports that shadowed the module-level traceback import and masked the real streaming exception with cannot access local variable 'traceback'.
- Wrapped direct source-backed queueing so descriptor/credential failures log [TABULAR_GENERATED_OUTPUT] Direct source-backed generated output queueing skipped and fall back to existing tabular analysis instead of aborting the stream.
- Added regression coverage for non-fatal direct queue fallback.

Validation rerun:
- .\\.venv\\Scripts\\python.exe -m py_compile application\\single_app\\route_backend_chats.py application\\single_app\\config.py functional_tests\\test_tabular_row_orchestration_scale.py
- .\\.venv\\Scripts\\python.exe functional_tests\\test_tabular_row_orchestration_scale.py

Issue #1031, opened 2026-07-14
Scale per-row tabular analysis and generated CSV exports to 30,000+ rows


paullizer:
## Issue
Per-row tabular analysis works for small files but fails to produce the requested exhaustive CSV once the source rows are split across multiple tabular tool pages. The customer scenario asks nine independent questions against every CSV row, requires one output row per source row, then expects an overall summary and any remaining prompt actions.

The attached reproduction succeeds with 10 rows and fails with 300 rows. The design must support 300, 3,000, and 30,000 rows with bounded memory and durable progress; larger files may take longer, but row count must not change correctness.

## Steps to Reproduce
1. Upload `priv-simplechat-2/tabular_issue/simplechat_row_orchestration_dataset_300.csv`.
2. Select the document in chat.
3. Ask SimpleChat to answer nine questions for each row, create a CSV with one output row per source row, then provide an overall analysis.
4. Compare the generated artifact and backend diagnostics with the 10-row fixture `simplechat_row_orchestration_dataset.csv`.

## Expected Behavior
- Every source row is processed exactly once and remains in source order.
- The generated CSV contains exactly one output row per source row with a stable schema and source identity.
- Work proceeds in bounded, checkpointed batches and can resume after worker or model-service interruption.
- The final CSV is committed and attached only after all batches validate.
- The assistant then returns a compact overall analysis and completes remaining requested actions.
- Progress, cancellation, retry, and failure details remain visible for long-running jobs.

## Actual Behavior
- The 300-row `query_tabular_data` result is safely paged into `94 + 95 + 94 + 17` rows.
- Generated-output source selection evaluates each invocation independently instead of treating compatible pages as one exhaustive result.
- The largest page reports `95` returned rows out of `300` total, so export creation logs `Selected source candidate is incomplete; skipping export`.
- The final assistant response may contain a small summary table; the generic assistant-table exporter then saves that summary as a misleading nine-row CSV rather than the requested 300-row deliverable.

## Root Cause
1. `route_backend_chats.py::_build_tabular_generated_output_source_candidate(...)` ranks individual tool invocations and requires one invocation to contain the full result.
2. Tool pagination from `TabularProcessingPlugin._build_tabular_row_page_payload(...)` is correct, but generated-output selection does not coalesce or replay those pages.
3. Durable background export queueing begins only after all source rows are already present and staged as one input-batches blob.
4. Background finalization loads all checkpoint outputs into one Python list before serializing and uploading the artifact.
[...]

PR #1178, opened 2026-08-09
Fix and accelerate LLM tabular background runs

paullizer:
## Summary - Fixes streaming direct source-backed tabular queue calls by passing required model and settings arguments with explicit keywords. - Automatically retries LLM row-count/schema/source-token validation failures, then preserves manual Continue from durable checkpoints after automatic retries are exhausted. - Keeps every row-level answer LLM-generated while increasing independent model-call concurrency through adaptive 4/16/64/128 tiers; explicit administrator concurrency still overrides the tier. - Sizes source batches from selected-model metadata or optional input/output token-limit fields in the local model capability catalog, with bounded fallback budgets, 50% input and 60% output targets, and a 180k soft input cap for contexts over 500k tokens. - Calculates ETA from recent wall-clock rows/minute across parallel windows instead of summing concurrent request durations as serial work. - Persists progress once per completed parallel window and exposes throughput, configured concurrency, and effective concurrency on the chat status card. - Bumps the application to 0.250.136 and updates feature/test documentation.  ## Production evidence - The 30,000-row source staged as 909 batches of about 33 rows under the prior 60k-character budget. - Runtime used concurrency 3 and reported about 35 hours remaining because concurrent request durations were serialized in ETA math. - At the observed roughly 155 seconds per 33-row model call, 100-128 effective asynchronous calls project the same LLM-only workload into roughly the 18-25 minute range before validation retries. Actual time remains model- and output-volume-dependent.  ## Validation - `python -m py_compile application/single_app/config.py application/single_app/functions_tabular_generated_exports.py functional_tests/test_tabular_row_orchestration_scale.py ui_tests/test_chat_background_generated_export_status.py` - `python functional_tests/test_tabular_row_orchestration_scale.py` - `node --check application/single_app/static/js/chat/chat-messages.js` - `python -m json.tool application/single_app/static/json/model_capabilities.json` - `python -m pytest ui_tests/test_chat_background_generated_export_status.py -q` (5 skipped locally because authenticated UI target variables are not configured) - VS Code diagnostics: clean for all changed code/test files - `git -c core.whitespace=blank-at-eol,blank-at-eof,space-before-tab,cr-at-eol diff --check`

Refs #1031

### DIFF
--- a/application/single_app/config.py
+++ b/application/single_app/config.py
@@ -96,7 +96,7 @@ DOTENV_LOAD_RESULT = load_simplechat_dotenv()
 EXECUTOR_TYPE = 'thread'
 EXECUTOR_MAX_WORKERS = 30
 SESSION_TYPE = 'filesystem'
-VERSION = "0.250.134"
+VERSION = "0.250.135"
 IS_DEVELOPMENT = is_development_env_enabled()
 
 SESSION_COOKIE_SAMESITE = os.getenv('SESSION_COOKIE_SAMESITE', 'Lax')
--- a/application/single_app/config.py
+++ b/application/single_app/config.py
@@ -96,7 +96,7 @@ DOTENV_LOAD_RESULT = load_simplechat_dotenv()
 EXECUTOR_TYPE = 'thread'
 EXECUTOR_MAX_WORKERS = 30
 SESSION_TYPE = 'filesystem'
-VERSION = "0.250.133"
+VERSION = "0.250.134"
 IS_DEVELOPMENT = is_development_env_enabled()
 
 SESSION_COOKIE_SAMESITE = os.getenv('SESSION_COOKIE_SAMESITE', 'Lax')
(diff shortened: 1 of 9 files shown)


paullizer:
Added retry capability for failed tabular generated exports:
- Model-output validation failures now auto-requeue up to 	abular_generated_output_model_validation_auto_retries (default 3).
- After the auto budget is exhausted, the run is marked failed but remains retryable so the existing manual Continue action can resume from checkpoints.
- Scheduler auto-claim skips exhausted failures; manual Continue clears the exhausted state and queues another attempt.
- Added regression coverage for auto retries followed by manual-only continuation.

Validation rerun:
- .\\.venv\\Scripts\\python.exe -m py_compile application\\single_app\\functions_tabular_generated_exports.py application\\single_app\\functions_settings.py application\\single_app\\config.py functional_tests\\test_tabular_row_orchestration_scale.py
- .\\.venv\\Scripts\\python.exe functional_tests\\test_tabular_row_orchestration_scale.py

Issue #1272, opened 2026-08-18
Completed Data Management jobs still show 'Current container: Waiting' and 'running' step badges


paullizer:
## Summary

On a finished Data Management backup job, the details panel still shows **Current container: Waiting**, and the timeline badges for completed steps (`Cosmos`, `Ai Search`, `Source Blobs`) still read `running`. Both are cosmetic, but they make a completed job look stuck.

### Stale "Current container"

`_execute_backup_source_blob_resource` builds progress with:

```python
"in_flight_count": len(current_files),
"current_file": sorted(current_files)[0] if current_files else "",
```

The final checkpoint is written after the worker pool drains, so `current_file` is `""` and the UI renders its empty-state label "Waiting" — on a job that is already `completed_with_warnings`.

### Timeline badges stuck on "running"

`_set_job_progress` defaults to `status=DATA_MANAGEMENT_STATUS_RUNNING` and stamps that onto the timeline event it records:

```python
def _set_job_progress(job, message, completed_steps, total_steps,
                      current_step=None, status=DATA_MANAGEMENT_STATUS_RUNNING, ...):
    ...
    _record_data_management_job_event(saved_job.get("id"), current_step or "progress",
                                      saved_job, status=status, message=message, ...)
```

This is historically accurate — the job *was* running when each step finished — but the UI presents the badge in a way that reads as current state.

## Expected behavior

- When a job reaches a terminal status, the details panel should not show a live "Current container" value; it should be hidden or show a completion summary.
- Timeline step badges should be visually distinguishable from live job status, or reflect the step outcome rather than the job status at event time.

## Notes

Observed on job `data_management_partial_20260818T0300Z` (SimpleChat `0.250.217`). No data impact; presentation only.

Related: #1258, #1271

PR #1273, opened 2026-08-18
Fix source blob backup ETag comparison failing every blob

### LINKED ISSUE #1271: Source blob backups fail 100% with 'Source blob changed while it was being backed up' (ETag format mismatch)
## Summary

Every source blob in every Data Management backup fails with `Source blob changed while it was being backed up.` The failure rate is exactly 100% across all source blob containers, with zero retries, so **no user document, group document, public document, or chat attachment has ever been backed up**.

The job still reports `completed_with_warnings`, which makes the condition easy to miss.

## Impact

- **Severity:** P0 — silent, total loss of source blob coverage in backups.
- **Blast radius:** every deployment with Enhanced Citations source blobs enabled.
- Also wastes substantial time and egress: the check runs *after* the blob is downloaded, encrypted, and uploaded, so the full transfer cost is paid and then discarded.
- Leaves orphaned `pending` artifacts in the backup container on every run.

## Observed behavior

From job `data_management_partial_20260818T0300Z`:

| Container | Blobs read | Copied | Failed |
|---|---|---|---|
| `user-documents` | 71 | 0 | 71 |
| `group-documents` | 427 | 0 | 427 |
| `public-documents` | 95 | 0 | 95 |
| `personal-chat` | 19,394 | 0 | 19,394 |

```
[DATA_MANAGEMENT] Source blob backup completed with failures. -- {'job_id': 'data_management_partial_20260818T0300Z',
 'resource': 'source_blobs:personal-chat', 'container': 'personal-chat', 'failed_count': 19394,
 'copied_count': 0, 'skipped_count': 0, 'source_read_count': 19394,
 'failure_reasons': '19394x Source blob changed while it was being backed up.'}
```

Job telemetry reports `Retries / throttles: 0 / 0`, confirming a non-transient first-attempt failure.

## Root cause

`_transfer_backup_source_blob` compares two ETags that come from different Azure SDK code paths with different transport formats:

```python
current_source_properties = source_blob_client.get_blob_properties()
current_source_etag = _safe_text(_get_backup_blob_property(current_source_properties, "etag"))
if source_etag and current_source_etag != source_etag:
    raise RuntimeError("Source blob changed while it was being backed up.")
```

| Value | Origin | SDK path | Format |
|---|---|---|---|
| `source_etag` | `list_blobs()` | `get_blob_properties_from_generated_code()` reads the XML `<Etag>` element | `0x8DE...` |
| `current_source_etag` | `get_blob_properties()` | `BlobProperties(**headers)` reads the HTTP `ETag` header | `"0x8DE..."` |

The HTTP `ETag` header is an RFC 7232 quoted-string; the List Blobs XML element is not. `azure-storage-blob==12.24.1` performs no normalization in either direction, so the comparison is always unequal.

`RuntimeError` is not retryable under `_is_retryable_backup_blob_error`, so every blob fails on the first attempt.

Independent confirmation that the blobs did not actually change: the ranged reads send that same unquoted ETag as an `If-Match` precondition and Azure accepts them. Only the Python-side string comparison is wrong.

## Secondary problem: one Cosmos write per blob

`record_transfer_result` calls `persist()` unconditionally for every item, producing one Cosmos checkpoint write per blob — 19,394 writes for `personal-chat` alone. This caps throughput at roughly 6 items/second and accounts for the 74-minute runtime. It will remain a bottleneck once blobs start succeeding.

## Expected behavior

- Source blobs are backed up successfully.
- A genuine mid-transfer source change is still detected and rejected.
- Checkpoint writes are batched so throughput is not bound by a Cosmos round-trip per item.

## Proposed fix

- Normalize both operands before comparing: strip transport quoting and the optional `W/` weak-validator prefix. Keep `source_item["source_etag"]` exactly as returned by `list_blobs()` so the `If-Match` precondition sent on ranged reads is unchanged.
- Batch checkpoints on manifest batch size (100) or a 15 second interval, whichever comes first, while still asserting the job lease on every item.

The migration path has a visually similar comparison but sources both operands from `get_blob_properties()`, so it is not affected.

## Validation

New functional test `functional_tests/test_data_management_backup_source_blob_etag.py` exercises the real transfer path with an unquoted listing and a quoted fetch. Neutralizing the normalization reproduces the exact production message:

```
Transfer must succeed, got 'failed' ('Source blob changed while it was being backed up.')
```

## Environment

- SimpleChat `0.250.217`
- `azure-storage-blob==12.24.1`
- Azure App Service, 3 instances

paullizer:
Fixes #1271

## Problem

Every source blob in every Data Management backup failed with `Source blob changed while it was being backed up.` — 100% failure rate, zero retries, across all containers. **User documents, group documents, public documents, and chat attachments have never been backed up.** The job still reported `completed_with_warnings`, which made it easy to miss.

From job `data_management_partial_20260818T0300Z`:

| Container | Blobs read | Copied | Failed |
|---|---|---|---|
| `user-documents` | 71 | 0 | 71 |
| `group-documents` | 427 | 0 | 427 |
| `public-documents` | 95 | 0 | 95 |
| `personal-chat` | 19,394 | 0 | 19,394 |

```
[DATA_MANAGEMENT] Source blob backup completed with failures. -- {'container': 'personal-chat',
 'failed_count': 19394, 'copied_count': 0, 'source_read_count': 19394,
 'failure_reasons': '19394x Source blob changed while it was being backed up.'}
```

## Root cause

`_transfer_backup_source_blob` compares two ETags that come from different Azure SDK code paths with different transport formats:

| Value | Origin | SDK path | Format |
|---|---|---|---|
| `source_etag` | `list_blobs()` | `get_blob_properties_from_generated_code()` reads the XML `<Etag>` element | `0x8DE...` |
| `current_source_etag` | `get_blob_properties()` | `BlobProperties(**headers)` reads the HTTP `ETag` header | `"0x8DE..."` |

The HTTP `ETag` header is an RFC 7232 quoted-string; the List Blobs XML element is not. `azure-storage-blob==12.24.1` normalizes neither — I grepped the installed package to confirm. So the check was always `0x8DE... != "0x8DE..."`.

`RuntimeError` is not retryable under `_is_retryable_backup_blob_error`, so every blob failed on the first attempt — matching the observed `Retries / throttles: 0 / 0`.

Independent confirmation the blobs did not actually change: ranged reads send that same unquoted ETag as an `If-Match` precondition and **Azure accepted them**. Only the Python-side comparison was wrong. The guard also runs *after* staging and commit, so every blob was fully downloaded, encrypted, and uploaded before being discarded — and left behind as an orphaned `pending` artifact.

## Changes

**1. ETag normalization** — new `_normalize_backup_etag` strips transport quoting and the optional `W/` weak-validator prefix, applied only at the comparison site.

`source_item["source_etag"]` deliberately keeps the exact value returned by `list_blobs()`, so the `If-Match` precondition sent to Azure is **byte-for-byte unchanged** from current production behavior. The guard is preserved — a genuine mid-transfer change still fails.

**2. Checkpoint batching** — `record_transfer_result` previously called `persist()` per item: 19,394 Cosmos writes for one container, capping throughput near 6 items/sec and stretching the run to 74 minutes. New `maybe_persist` checkpoints on manifest batch size (100) or a 15-second interval, whichever comes first. The job lease is still asserted every item, and the existing tail `persist()` still flushes the final partial batch. Worst-case re-work after an interrupted run stays bounded at 100 items or 15 seconds.

**3. Functional tests that silently passed under pytest** — while validating this fix I found the repo's standard test template hides failures:

```python
def test_x():
    try:
        assert ...
        return True
    except Exception:
        return False
```

Returning a value instead of raising makes pytest report the test as **passed** with only a `PytestReturnNotNoneWarning`. A deliberately broken build reported `7 passed` while the transfer was genuinely failing. Both backup test files now assert directly and return `None`, with the `__main__` block preserving standalone output and exit codes. Suite warnings dropped 13 → 1.

## Scope check

The migration path (`_copy_source_blobs_to_target`) has a visually identical comparison, but sources `source_properties` from `get_blob_properties()`, so both operands are already quoted. **Migration is not affected** and is left unchanged.

## Validation

New test `functional_tests/test_data_management_backup_source_blob_etag.py` drives the real `_transfer_backup_source_blob` against in-memory blob clients with an unquoted listing and a quoted fetch.

```
7 passed in 2.51s
```

| Test | Assertion |
|---|---|
| `test_etag_normalization_strips_transport_quoting` | Quoted, unquoted, weak, padded ETags normalize identically |
| `test_listed_and_fetched_etags_compare_equal` | Raw listed ETag preserved for `If-Match`; both formats equal once normalized |
| `test_transfer_succeeds_across_list_and_get_etag_formats` | End-to-end transfer succeeds and promotes metadata to `succeeded` |
| `test_genuinely_changed_source_blob_still_fails` | ETag changing after download is still rejected |
| `test_verified_artifact_matches_source_version` | Reuse detection still keys off recorded source version |
| `test_checkpoint_interval_is_bounded` | Interval and batch size within safe bounds |
| `test_version_is_at_least_fix_version` | Version floor |

**Regression probe:** neutralizing `_normalize_backup_etag` fails 4 of 7 tests, including the end-to-end transfer, reporting the exact production message:

```
Transfer must succeed, got 'failed' ('Source blob changed while it was being backed up.')
```

**Full Data Management suite:** 154 passed, 1 failed. That failure (`test_backup_recovery_and_admin_progress_are_bounded_and_sanitized`) was confirmed pre-existing on `origin/Development` in the previous PR and is unrelated.

## Version

`0.250.217` → `0.250.218`

## Follow-up

#1272 tracks the cosmetic issue where completed jobs still display `Current container: Waiting` and `running` step badges.

## Documentation

- `docs/explanation/fixes/SOURCE_BLOB_BACKUP_ETAG_FIX.md`
- `docs/explanation/release_notes.md`


paullizer:

paullizer:

PR #1297, opened 2026-08-19
Admin Settings: move Latest Features to the bottom and split the settings template

paullizer:
First delivery stage of the Admin Settings information architecture rework.

> **Targets the `feature/admin-settings-ia` staging branch, not `Development`.** Nothing here reaches `Development` until the full IA rework is finished and tested, at which point the staging branch is merged in one reviewed PR.

## Why

Admin Settings had grown organically into 17 flat tabs in a single 13,526-line, 1 MB template. The most visible symptom was that **Latest Features opened on every single visit** — a curated release-notes tab behaving like the landing page.

### Latest Features no longer hijacks the page

Three separate places forced this and all three are fixed:

| Location | Before |
|---|---|
| Top-tab strip | First `<li>`, `class="nav-link active"` |
| Tab pane | `class="tab-pane fade show active"` |
| `admin_sidebar_nav.js` | Bootstrapped to `showAdminTab('latest-features')` |

Latest Features now sits **last in both navigations**, after Send Feedback, and **General is the landing tab**. Its content, New badge, and hide/unhide option are untouched.

### Template split into per-tab partials

Each tab pane moved verbatim into `templates/admin/_panes/` and is pulled back in with `{% include %}`. The parent keeps the global `<form>`, the modals, and the script blocks, and drops from 13,526 lines to about 2,200.

### Two features that were rendered as afterthoughts

- **Global Identities** was a bare JavaScript mount with no heading or description. It now has both, and explains that secrets go to Key Vault when Key Vault storage is configured.
- **File Sync** — ~400 lines, 32 inputs — was the only tab with no sidebar submenu, so its sub-areas couldn't be reached or searched. Its source-type and per-workspace-type areas are now sidebar destinations.

### Bugs found along the way

- **Classification banner preview never worked.** Its script sat between `{% endblock %}` and `{% block scripts %}`, and Jinja discards content outside blocks in a child template, so it was never rendered.
- **Home Page Text preview reinterpreted editor text as HTML.** The fallback assigned a textarea's raw `.value` to `innerHTML` despite the comment saying "just show raw text" (CodeQL `js/xss-through-dom`, high). Now `textContent`, with the Markdown branch sanitized inline via `DOMPurify.sanitize(...)` at the sink.
- **Sidebar `sectionMap` had rotted.** 72 entries, of which 66 mapped a key to itself (already handled by the `|| sectionId` fallback), two were unreferenced, and `control-center-admin-section` pointed at an element that doesn't exist. Now 6 real aliases, with a test preventing regrowth.

### CI coverage for the staging branch

Every gate in this repo is pinned to `branches: [Development]`, so PRs into an integration branch would have run **only** `enforce-branch-flow` — no CodeQL, no XSS sink check, no access-control or syntax checking. The final commit adds `feature/admin-settings-ia` to the `pull_request` filters of six gates so staged work is checked exactly like a `Development` PR. `codeql.yml`'s `push` trigger is deliberately unchanged, and `release-notes-check` is excluded because it already fails on fork PRs for token reasons and is non-blocking.

**This is reverted in the final staging → `Development` PR**, so `Development`'s CI configuration ends up untouched.

## Verification

**The save payload cannot have changed.** Every `name=` attribute and card id was fingerprinted before any edit and compared after each stage, and again after rebasing onto current `Development`:

```
OK   field_names: 462 identical
OK   card_ids: 110 identical
OK   field_name_counts: 452 unique names, counts unchanged
```

**Zero test regressions.** All 75 functional test files referencing `admin_settings.html`, run against the `Development` baseline and after every stage: **33 failures before, 33 after, identical sets.** Those 33 are pre-existing and unrelated (workflow route registration, Cosmos document-access wiring, Send Feedback docs).

All 20 admin templates parse under Jinja. `scripts/check_xss_sinks.py` and `functional_tests/test_xss_guardrails_checker.py` both pass locally. All 11 CI checks pass.

### Rebase note

`Development` moved 14 commits during this work, including a new
`shared-conversation-file-approvals-section` card added to the region this PR split into partials. That card was ported into `admin/_panes/ai-models.html` in the position upstream placed it, which is why the fingerprint is 462/110 rather than the 461/109 quoted earlier. Version renumbered to `0.260.007` to clear upstream's `0.260.004`–`0.260.006`.

## Note for reviewers

Tests that read `templates/admin_settings.html` directly now see only the parent shell. Use `test_support.templates.read_admin_settings_template()`, or `compose_if_admin_settings()` inside a shared file-reading helper.

The new `test_admin_settings_template_composition.py` enforces this — it fails if a test references a partial-backed card **or form field** while reading the template uncomposed. It caught 10 such tests, including 3 that were passing only because the assertion happened to be satisfied by the parent shell.

## Still to come, into the same staging branch

- **A:** nav map as a single source of truth for the IA; convert the 12 hardcoded `switchTab` cross-tab links to card-targeted links that survive renames
- **B:** declarative dependency gating — inline mirror plus a resolved link when a setting needs another one enabled; unify the 10 hand-written Latest Features mirrors
- **C:** group navigation level, three-level search, legacy hash redirect map
- **D:** re-home cards into the 14 groups, one group per commit
- **E:** split `system-settings-section`; consolidated Access and Roles roster

Fix documentation in `docs/explanation/fixes/ADMIN_SETTINGS_NAVIGATION_AND_TEMPLATE_STRUCTURE_FIX.md`.


paullizer:

PR #1305, opened 2026-08-19
Restrict app-identity action endpoints to Azure origins

paullizer:
## Summary

Actions that authenticate with the application's own workload identity could be configured with a caller-supplied endpoint that was never checked. This change restricts those endpoints to canonical Azure service origins, enforces each action type's declared authentication methods on the server, and constrains Log Analytics custom clouds.

Reported through MSRC. Details are kept general here deliberately.

## What changed

**New shared allowlist** — `functions_azure_endpoint_validation.py` holds hardcoded Azure host allowlists for Blob, Queue, Cosmos DB, Databricks, Azure Monitor query endpoints, and Microsoft Entra authority hosts, covering the public, US Government, China, and Germany clouds. Validators reject non-HTTPS schemes, embedded credentials, explicit ports, query strings, fragments, local hostnames, and IP literals, then return a canonical rebuilt origin.

**Enforced at two points**
- Save time, in `PluginHealthChecker.validate_plugin_manifest` — the shared choke point for personal saves, group create/patch, admin add/edit, every `test-*-connection` route, and MCP discovery.
- Runtime, immediately before client construction or token acquisition in each plugin, so actions stored before this change stop working rather than continuing to use the application identity against an unchecked destination.

**`allowedAuthTypes` is now enforced.** Each `<type>.definition.json` declares supported auth methods, but the value was previously only used to populate the action modal. `GET /api/plugins/<type>/auth-types` now resolves through the same helper so the UI and backend cannot drift.

**Log Analytics custom clouds** — `authorityHost` must be a known Entra authority and `endpointOverride` a known Azure Monitor query endpoint. The override also feeds the delegated-token OAuth scope, so both are constrained.

**Cosmos test-connection route** builds its own client and is validated independently.

**Blob storage UI** — `blob_storage.definition.json` widened to `connection_string`, `identity`, `key`, and the modal gained an auth selector plus endpoint and account-key fields, with a client-side mirror of the hostname allowlist. Previously the modal only collected a connection string, so managed identity was not reachable from the UI.

**File Sync refactor** — `functions_file_sync.py` now sources the storage suffix allowlist from the shared module instead of its own copy added in v0.250.068.

## Compatibility

Deployments using standard Azure hostnames are unaffected. Private Endpoint setups continue to work via the standard account hostname with private DNS. Custom domains, Azurite/development storage, Azure Stack, and direct private-link hostnames are intentionally rejected — same policy as the Azure Blob File Sync hardening in v0.250.068.

One behavioural note: `allowedAuthTypes` enforcement applies only to caller-declared auth types and **exempts identity-bound manifests**. Identity-bound actions have `auth.type` rewritten server-side during hydration, and the connection-test path validates after that rewrite, so enforcing there would reject valid configurations. Every definition file was checked against what the UI and hydration actually produce before enabling this.

## Testing

New `functional_tests/test_action_app_identity_endpoint_hardening.py` — 10/10 passing. Covers accepted canonical endpoints per cloud, rejection of unapproved/internal/metadata-service/lookalike/credential-bearing/port-bearing/query-bearing/fragment-bearing endpoints, cross-service mismatches, auth-type enforcement including the identity-bound exemption and the legacy Databricks type, Log Analytics custom-cloud constraints, runtime rejection of stored manifests, modal wiring, and the File Sync refactor.

Also re-ran: `test_file_sync_azure_blob_storage.py`, `test_mcp_destination_governance_and_preconfigurations.py`, and all three `route_tests/` suites — passing.

Two failures were confirmed pre-existing on the unmodified branch via `git stash` and are not introduced here: `test_file_sync_routes_do_not_disclose_exception_details`, and `test_blob_storage_action_capabilities.py`, which requires live Azure credentials to import.

## Follow-ups (not in this PR)

- A non-administrator can still create a personal `managed_identity` workspace identity and bind it to an action. That no longer permits arbitrary-origin credential delivery, but restricting such identities to admin/global scope is worth a separate decision.
- The bundled Bicep grants the web app built-in ARM `Contributor` on the Cosmos DB account, which widens the blast radius of any leaked token. Least-privilege review is a separate change — no `deployers/` files are touched here, so `deployers/version.txt` is not incremented.
- The upstream SDK behaviour that prompted the report needs a fix in `Azure/azure-sdk-for-python` across the Blob, Queue, File Share, and Data Lake challenge policies. Nothing in this PR substitutes for that.

Version bumped to `0.260.006`. Fix documentation in `docs/explanation/fixes/ACTION_APP_IDENTITY_ENDPOINT_HARDENING_FIX.md`.


paullizer:

PR #1432, opened 2026-09-04
Add chat orchestration framework (V2)

paullizer:
Adds the chat orchestration framework, and fixes the three symptoms found while testing it on a live deployment.

## What this adds

Lets a user describe what they want instead of assembling the request. Instead of choosing documents, search scope, web search, a saved prompt, an agent and a model before asking, the user asks; SimpleChat plans the work, shows the plan, and runs it once approved.

The planner model never invokes anything. It is shown a short list of declarative capability descriptors and returns schema-constrained JSON; a deterministic executor then dispatches through adapters over the retrieval and analysis functions that already exist. This is deliberately not Semantic Kernel function calling, which with 44 plugin classes and 27 actions is neither reliable nor cheap. Semantic Kernel remains how an agent runs.

Planner output is treated as untrusted input. A plan naming an unknown or disabled capability, referencing an unreadable document, or containing a dependency cycle is caught before an adapter is reached, and repairs are recorded rather than applied silently. Authorization is checked twice, at plan time and again before the answer is composed, because access can be revoked between them.

Questions are asked inline as a JSON-Schema form card shaped to the MCP elicitation specification, so a future MCP server reuses it. Plan detail and live progress live in a new plan mode of the right-hand drawer, with a Run view for the current plan and a Map view of the conversation's runs over time. That map is also a planner input: a bounded run ledger lets later turns reuse earlier findings instead of repeating them, and stops an elicitation asking the same question twice.

`route_backend_chats.py` is untouched.

## Second commit: making it reachable

Testing on a live deployment produced three symptoms -- no Orchestrate toggle, no approval control, no planning UI -- which all traced to one cause. The SPA reads its configuration once at page load, and Admin Settings is a different page. Enabling orchestration there left an already-open chat tab holding a payload from before the change, so the composer correctly concluded the feature did not exist and every downstream control stayed hidden. Nothing on screen explained why.

Bootstrap is now re-read when the tab returns to the front. Orchestration also defaults on wherever an administrator has enabled it, with the manual controls folded behind a disclosure rather than removed.

## Verification

- 23 functional tests (registry contract, plan schema, MCP elicitation, run ledger, executor including re-authorization)
- 31 Playwright UI tests (plan card, countdown, auto-open asymmetry, narrowing-only editing, elicitation paging, Run/Map, composer)
- 12 route policy tests, 13 docs tests
- 47 regression tests green on the shared admin and bootstrap surfaces
- typecheck, build, swagger validation and pyflakes all clean

## Scope

Off by default, V2 only, classic interface unchanged. Retrieval and reasoning capabilities in this phase; image generation, exports, agent dispatch and per-step model routing are deferred, the last because the model catalogue records no capability metadata to route on.


paullizer:
Superseded by #1433, which is the same branch on microsoft/simplechat rather than a cross-fork copy. Direct push was briefly unavailable when this was opened.

PR #1454, opened 2026-09-06
Hydrate orchestration run history so plans survive a reload or a device change

paullizer:
## The report

> the orchestration plans seem to stay on the client side and are not retained as part of the conversation, if i switch devices i dont see previously ran plans started on other devices...is this real or is something else happening?

It was real. Nothing was lost, though.

## What was actually happening

Runs, plans and steps have always been persisted server-side — `cosmos_orchestration_runs_container` (partitioned by `/conversation_id`) and `cosmos_orchestration_run_steps_container` (partitioned by `/run_id`) — and the planner already reads earlier runs back to build its ledger.

The V2 interface just never asked for them:

- Nothing under `application/v2_ui/src` referenced `/api/v2/orchestration/runs` or `/runs/<id>/steps`. Two endpoints existed and were called by nothing.
- `restorePersistedRuns()` was exported from the orchestration store and invoked nowhere, so even a same-device reload lost the history.
- There was no route returning a single run *with its plan*, so the panel had no way to fetch a plan it had not itself produced.
- The saved user message did not reliably carry its `orchestration_turn_id`, so a hydrating client could not line a stored run up with the message that caused it.

The drawer ran entirely off in-memory Zustand state backed by `sessionStorage`, which is per-tab and per-device by definition.

## Behaviour change

| Situation | Before | After |
| --- | --- | --- |
| Open a conversation on a second device | Empty panel | Earlier runs listed, expandable to steps and results |
| Reload the page mid-conversation | Run history lost | Rebuilt from the server |
| Plan left awaiting approval, user moves device | Stranded; the question had to be asked again | Reappears as a plan that can be approved, edited or discarded |
| That plan was already approved on the first device | Would race | Refused, and the existing answer is loaded |
| Browser closed mid-run | No record | Listed as interrupted |

## Server

- **New** `GET /api/v2/orchestration/runs/<run_id>` returning one run with its plan. Ownership is enforced inside `get_orchestration_run` on both the point read and the cross-partition fallback, so a guessed run id belonging to someone else is indistinguishable from a missing one.
- **Allowlist projections** (`_run_summary_row`, `_run_detail_row`) decide what leaves the server. Listing 25 runs no longer puts 25 full plans on the wire, and `seeds`, `user_id`, `approved_by`, `conversation_context`, `request_resolution`, `user_message_fingerprint`, Cosmos system fields and raw artifact payloads all stay server-side. This replaces a blocklist that stripped three of those by name — an allowlist keeps excluding whatever the record gains later. `?include_plan=true` *widens* the projection rather than bypassing it, so no caller reaches a stored record verbatim.
- **`orchestration_turn_id` is stamped on every orchestrated question.** `_save_turn_message` already wrote it, but only when the turn happened to carry a selected prompt — so the flat key was absent on ordinary questions even though `chatStore.ts` already read it.

## Client

- Typed read layer: `fetchConversationRuns`, `fetchOrchestrationRun`, `fetchRunSteps`.
- Store hydration merges persisted summaries into history, local runs winning over hydrated ones. A restored run the server reports as finished is dropped from the in-flight set instead of spinning forever.
- A run opened from history is read-only, with a banner saying it is a record.
- A plan still awaiting approval is restored as approvable, guarded three ways: newest run only, nothing already active locally, and no assistant message already following the question.
- A restored **timed** approval is neutralised to manual, so a countdown started on one device cannot run real work unattended on another.
- The run route's existing HTTP 409 re-run guard is now handled client-side: the local turn settles and the conversation reloads rather than surfacing an error.

## No migration

Run records already carried `turn_id` and `user_message_id`. Conversations predating this hydrate correctly; every path using the new stamp falls back to `user_message_id`.

## Validation

- `functional_tests/test_orchestration_run_hydration_routes.py` — 6/6
- `functional_tests/test_orchestration_turn_recovery_identifiers.py` — 6/6
- `ui_tests/test_v2_orchestration_run_hydration.py` — 6/6
- `ui_tests/test_v2_orchestration_pending_approval_resume.py` — 5/5
- All `test_v2_orchestration_*` UI suites, all three `route_tests/` policy suites, and both docs suites pass
- `tsc -b --noEmit` and `npm run build` clean; docs inventory regenerated (unchanged — no new settings, tabs, actions or chat controls)

**Pre-existing failure, not from this branch:** `test_approve_runs_the_plan` in `ui_tests/test_v2_orchestration_plan_card.py` fails 6/7. I verified it fails identically on a clean checkout of the base at `7341323c`, so it is untouched here.

## Notes for review

- Base is **`paullizer-react-v2-ui`**, not `main`.
- Opened from a fork because push access to `microsoft/simplechat` was lost partway through the session. The branch `paullizer-orchestration-plan-persistence` pushed there earlier is **stale pre-rebase content** and should be ignored or deleted — this PR's head has the rebased, conflict-resolved work.
- Version `0.261.098`. Docs: `docs/explanation/fixes/ORCHESTRATION_RUN_HISTORY_HYDRATION_FIX.md`, a run-history section plus troubleshooting rows in `docs/admin/orchestration.md`, and a release notes entry.

### DIFF
--- a/application/single_app/config.py
+++ b/application/single_app/config.py
@@ -97,7 +97,7 @@ DOTENV_LOAD_RESULT = load_simplechat_dotenv()
 EXECUTOR_TYPE = 'thread'
 EXECUTOR_MAX_WORKERS = 30
 SESSION_TYPE = 'filesystem'
-VERSION = "0.261.095"
+VERSION = "0.261.096"
 IS_DEVELOPMENT = is_development_env_enabled()
 
 SESSION_COOKIE_SAMESITE = os.getenv('SESSION_COOKIE_SAMESITE', 'Lax')
--- a/application/single_app/route_backend_orchestration.py
+++ b/application/single_app/route_backend_orchestration.py
@@ -130,6 +130,13 @@ def _text(value, limit=None):
     return value[:limit].rstrip() if limit and len(value) > limit else value
 
 
+def _coerce_int(value, default=0):
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return default
+
+
 def _sse(generator):
     return Response(generator, mimetype='text/event-stream', headers=dict(SSE_HEADERS))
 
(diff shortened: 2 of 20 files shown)


paullizer:
Superseded by #1455.

This PR was opened from a branch in this repo that holds pre-rebase commits and conflicts with \paullizer-react-v2-ui\. Push access to this repo was lost partway through the session, so the branch could not be updated with the rebased, conflict-resolved work. #1455 carries the same change from a fork and is mergeable.

